PR #7, opened 2019-09-22
Basic Linker

splashofcrimson:
Still to do:

1. Link files directly while calling bril2json
2. Scope functions within the modules they came from (to avoid multiple declarations)
3. Test cases


splashofcrimson:
Sorry! I made a pull request in the wrong repository

PR #16, opened 2019-09-26
Project 1: Function calls

avanhatt:
@gyauney and my implementation of Bril function calls, as [proposed][] and [described][] in detail elsewhere!

[proposed]: https://github.com/sampsyo/cs6120/issues/14
[described]: https://github.com/sampsyo/cs6120/pull/38


sampsyo:
Merged this function stuff in #43! Thank you for all this work so long ago!!

PR #20, opened 2019-09-26
Fixed Size Arrays

Checkmate50:
Implemented fixed size arrays with indexing and assignment in Bril, `bril2json`, `bril2txt`, `js2bril`, and `brili`.  Converting from javascript is highly limited and will produce unknown-sized arrays (length of `-1`), which needs to be fixed in a later pass.  Array size can be trivially inferred in a pass, since the only permitted operations that produce arrays are `new` and `id`.


sampsyo:
Yo! We now have heaps and allocation, so I'm closing this (worthy!) alternative plan.

PR #21, opened 2019-09-26
Cleanup speedup

wbthomason:
- Some general code cleanup
- Use indices and a vec instead of variable names and a hash
- Minor bugfixes


wbthomason:
Whoops, premature PR.

PR #26, opened 2019-10-02
Project 1: Shrimp - Symbolic Verification for Bril

sgpthomas:
A tool for verifying that optimization passes preserve the functional correctness of basic blocks. This is written in Rosette and provides a cli that simply takes in two Bril programs in either the `json` or `bril` format and either returns a counter example showing that two blocks differ or says that the two programs are functionally equivalent.

Shrimp assumes that both programs have the same basic blocks.

Sorry that this was late, we forgot to do this earlier.


sampsyo:
Yo; I am clearing out old PRs to spruce up Bril v2.0 for the fall. I hope you are enjoying your post-6120 life!

PR #27, opened 2019-10-03
Probabalistic Support and Exact Interpreter

orichardson:
- rand in interpreter
- new file `xbrili` which abstractly exactly calculates the distribution, including all limit fixed points, for finite state space, and also all branches of an infinite space that exceed a minimum probability mass, without relying on Jacobi iterates of increasing precision, or direct computation of the complete eigenspace.
- a tool for randomly generating programs
- some changes to the structure of the bril interpreter that will definitely need to be massaged for compatibility with other bril extensions


sampsyo:
Yo; I am clearing out old PRs to spruce up Bril v2.0 for the fall. I hope you are enjoying your post-6120 life!

PR #41, opened 2019-10-28
Update LVN to bail on division by zero (fixes #40)

### LINKED ISSUE #40: LVN should not crash on divide-by-zero
The current implementation of constant folding via local value numbering in `lvn.py` crashes when the program divides by zero. Instead of crashing, the compiler should bail on folding constants when there are dynamic errors. 

Minimal example:
```
void main() {
entry:
  zero : int = const 0;
  one : int = const 1;
  baddiv : int = div one zero;
  print baddiv;
}
```

Stack trace:
```
Traceback (most recent call last):
  File "examples/lvn.py", line 243, in <module>
    lvn(bril, '-p' in sys.argv, '-c' in sys.argv, '-f' in sys.argv)
  File "examples/lvn.py", line 236, in lvn
    fold=_fold if fold else lambda n2c, v: None,
  File "examples/lvn.py", line 160, in lvn_block
    const = fold(num2const, val)
  File "examples/lvn.py", line 211, in _fold
    return FOLDABLE_OPS[value.op](*const_args)
  File "examples/lvn.py", line 197, in <lambda>
    'div': lambda a, b: a // b,
ZeroDivisionError: integer division or modulo by zero
```

avanhatt:
Also add example testing to Makefile.

The current implementation of constant folding via local value numbering in lvn.py crashes when the program divides by zero. Instead of crashing, the compiler should bail on folding constants when there are dynamic errors.

### DIFF
--- a/examples/dom_test/loopcond.out
+++ b/examples/dom_test/loopcond.out
@@ -1,0 +1,1 @@
+{'entry': {'entry'}, 'loop': {'loop', 'entry'}, 'exit': {'loop', 'exit', 'entry'}, 'body': {'loop', 'entry', 'body'}, 'then': {'then', 'loop', 'entry', 'body'}, 'endif': {'loop', 'endif', 'entry', 'body'}}
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,19 @@ TESTS := test/parse/*.bril \
 	test/interp/*.bril \
 	test/ts/*.ts
 
+EXAMPLE_TESTS :=  examples/*_test/*.bril 
+
 .PHONY: test
 test:
 	turnt $(TESTS)
+
+.PHONY: test_examples
+test_examples:
+	turnt $(EXAMPLE_TESTS)
+
+.PHONY: test_examples_save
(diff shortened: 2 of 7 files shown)


examples/dom_test/loopcond.out
```diff
@@ -0,0 +1 @@
+{'entry': {'entry'}, 'loop': {'loop', 'entry'}, 'exit': {'loop', 'exit', 'entry'}, 'body': {'loop', 'entry', 'body'}, 'then': {'then', 'loop', 'entry', 'body'}, 'endif': {'loop', 'endif', 'entry', 'body'}}
```
avanhatt: The bummer about this one is that as implemented, `dom.py` doesn't produce a deterministic list. Our project 2 enforces CFG order on it with a somewhat hacky solution; but short of that the main Bril repo could also just sort based on something arbitrary like the block name.

sampsyo: Good point! Maybe that command should produce an OrderedDict or something… but in any case, seems fine for now!

PR #55, opened 2020-09-08
Add sieve benchmark

tedbauer:


sampsyo:
This is awesome! Looks like you coded this "by hand" in Bril, huh? Very cool implementation, and it's likely to be somewhat harder to optimize than some of the TypeScript-generated benchmarks we have elsewhere. 👏

PR #58, opened 2020-09-14
Add ackermann benchmark

kq-li:


sampsyo:
Great! A pretty good argument for adding tail calls to the interpreter…

PR #60, opened 2020-09-14
added sqrt benchmark

evanbenadler:
This benchmark uses Newton's Method to approximate square roots. I've included the typescript file in the benchmark directory but please feel free to remove it.


sampsyo:
Looks good, but can you please commit the text format instead of the JSON? (That should be what works with `turnt` in this directory.)

PR #64, opened 2020-09-14
Add Matrix Multiplication benchmark

ankushrayabhari:


ankushrayabhari:
In case you're wondering if the output is actually correct, I tested it with this python script.

```
import numpy as np

f = open("mat-mul.out", "r")
flat = np.loadtxt(f)
[aflat, bflat, cflat] = np.split(flat, 3)
a = np.reshape(aflat, (-1, 50))
b = np.reshape(bflat, (-1, 50))
c = np.reshape(cflat, (-1, 50))
np.testing.assert_equal(np.dot(a, b), c)
```

sampsyo:
Wahoo; awesome work!! 🙌

PR #69, opened 2020-09-14
Add gcd benchmark

hj424:
Greatest Common Divisor using Euclidean algorithm


sampsyo (approved):
Looks awesome; thanks!!

PR #73, opened 2020-09-15
add incredibly naive pythagorean triple finder

Calsign:


sampsyo:
This is a good start! As the task description says, can you please add to the docs and include the test output?

sampsyo (changes requested):
This is a good start! As the task description says, can you please add to the docs and include the test output?

Calsign:
Should be better now, sorry about that

sampsyo:
Looks good; thanks!!

PR #74, opened 2020-09-15
Add Digital Root Benchmark

EclecticGriffin:
Added a benchmark example that computes the digital root of a number and updates the markdown doc. 

Also because my editor is obnoxious it removed a trailing space on another line.


sampsyo (approved):
Excellent! And I'd call removing trailing whitespace altruistic, not obnoxious at all. :smiley:

PR #76, opened 2020-09-16
Adding opam support for bril-ocaml!

priyasrikumar:
Currently, users of bril-ocaml need to write their code within the library directory in order to use it.
This PR will allow users to run the command `opam pin add bril https://github.com/sampsyo/bril/tree/master/bril-ocaml` to use bril.

(also updated the docs to reflect this!)


bril-ocaml/bril.opam
```diff
@@ -0,0 +1,18 @@
...
+synopsis: "Bril wrapper"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "core"
+  "ocaml"
+  "yojson"
+  "ppx_jane"
```
sampsyo, line 16: I think [core](https://opam.ocaml.org/packages/core/) already includes ppx_jane, so this dependency can probably be omitted.

docs/tools/ocaml.md
```diff
@@ -12,7 +12,7 @@ Then, install the dependencies with `opam install core yojson`.
 Use
 ---
 
-You can include the library by adding it to the `libraries` substanza of your `dune` file.
+You can include the library by running `opam pin add bril https://github.com/sampsyo/bril/tree/master/bril-ocaml`. You can then include it in your dune files as `bril`, like any other library!
```
sampsyo, line 15: Maybe this would be more legible as a standalone code block, perhaps? Like this:

```
…by running:

    opam pin add bril https://github.com/sampsyo/bril/tree/master/bril-ocaml

You can then…
```

sampsyo (commented):
Looks great! Here are a couple of minor suggestions.

sampsyo:
Thanks!

PR #78, opened 2020-09-18
Update bril.ml

stp59:
There is a bug in the deparser that causes the types of void functions to be printed as `: None`. This should fix it.

### DIFF
--- a/bril-ocaml/lib/bril.ml
+++ b/bril-ocaml/lib/bril.ml
@@ -250,15 +250,27 @@ let to_string { funcs } =
     | Nop -> `Assoc [ ("op", `String "nop") ]
   in
   let of_func { name; args; ret_type; instrs; _ } =
-    `Assoc
-      [
-        ("name", `String name);
-        ( "args",
-          `List
-            (List.map args ~f:(fun (name, bril_type) ->
-                 `Assoc [ ("name", `String name); ("type", of_type bril_type) ])) );
-        ("type", Option.value_map ret_type ~default:`Null ~f:of_type);
-        ("instrs", `List (List.map instrs ~f:of_instr));
-      ]
+    if Option.is_some ret_type then
+      `Assoc
+        [
+          ("name", `String name);
+          ( "args",
+            `List
+              (List.map args ~f:(fun (name, bril_type) ->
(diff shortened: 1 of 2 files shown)


stp59:
I've also removed `Not` from the type for `binop` which was causing a bug.

bril-ocaml/lib/bril.ml
```diff
@@ -250,15 +248,27 @@ let to_string { funcs } =
...
+      `Assoc
+        [
+          ("name", `String name);
+          ( "args",
+            `List
+              (List.map args ~f:(fun (name, bril_type) ->
+                   `Assoc [ ("name", `String name); ("type", of_type bril_type) ])) );
+          ("instrs", `List (List.map instrs ~f:of_instr));
+        ]
+
```
kq-li, line 272: think we can shorten this?
```
`Assoc ([
  ("name", ...);
  ("args", ...);
  ("instrs", ...);
] @ (Option.value_map ret_type ~default:[] ~f:(fun ret_type -> [of_type ret_type])))
```

stp59: Nice!

bril-ocaml/lib/bril.ml
```diff
@@ -250,15 +248,13 @@ let to_string { funcs } =
...
-        ("instrs", `List (List.map instrs ~f:of_instr));
-      ]
+      `Assoc (
+        [ ("name", `String name);
+          ( "args",
+            `List
+              (List.map args ~f:(fun (name, bril_type) ->
+                   `Assoc [ ("name", `String name); ("type", of_type bril_type) ])) );
+          ("instrs", `List (List.map instrs ~f:of_instr));
+        ] @ (Option.value_map ret_type ~default:[] ~f:(fun t -> [of_ret_type t])))
```
kq-li, line 258: i think this is a typo, `s/of_ret_type/of_type/`

kq-li: just realized it would have to be `[("type", of_type t)]` -- i've included these fixes in #80

kq-li:
i included your fixes in #80 so we don't need to merge later!

PR #79, opened 2020-09-20
Add timeout option to brench

ankushrayabhari:


sampsyo (approved):
Looks great; thank you!! I just slightly simplified the default fallback for the config option.

PR #80, opened 2020-09-20
Modularize OCaml library

kq-li:


sampsyo (approved):
Looks great; thanks!!

PR #81, opened 2020-09-21
Fix basic block construction bug

kq-li:
The basic block construction algorithm didn't correctly terminate at labels.


sampsyo:
Good catch!

PR #83, opened 2020-09-21
ocaml library: include labels when converting to json

Calsign:
The implementation of `to_json` for the OCaml library is not including any labels, so it's basically broken. This fix adds the labels before each basic block. It also keeps track of "extra" labels which are not bound to any basic block, which happens when there is a label as the very last instruction.

Not thoroughly tested, but I am using this for my lesson 3 implementation so I thought I would make a PR.

On a related note, I think there can be multiple labels that point to a single basic block, which happens when there are two (or more) labels in a row. I think the current implementation will end up printing the same basic block twice in this case, which is wasteful but not necessarily wrong, but I haven't tested yet.

### DIFF
--- a/bril-ocaml/lib/func.ml
+++ b/bril-ocaml/lib/func.ml
@@ -8,10 +8,13 @@ type t = {
   blocks : Instr.t list String.Map.t;
   order : string list;
   cfg : string list String.Map.t;
+  extra_labels : string list;
 }
 [@@deriving compare, sexp_of]
 
-let instrs { blocks; order; _ } = List.concat_map order ~f:(Map.find_exn blocks)
+let instrs { blocks; order; extra_labels; _ } =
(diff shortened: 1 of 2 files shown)


Calsign:
@kq-li paging you since you made the OCaml library

kq-li:
good catch! i'm a bit unsure about whether the current behavior re: multiple labels in a row is correct. in 
```
l1:
l2:
  <instr>
```
maybe `l1` should be an empty block, and `l2` should have one instruction. if that is the case, we could also address this issue by actually adding the label "instruction" into the block `instr list` (this would also encapsulate the "extra labels" problem)

kq-li:
i opened #84 to encompass the above fix -- if you could take a look and let me know what you think that would be great (it should fix the issue you described too, hopefully)

PR #84, opened 2020-09-21
Fix label processing, again

kq-li:
Labels didn't make it to output JSON -- adding them as actual instructions in basic blocks should fix it. Also, double labels and trailing labels should now become "empty" (i.e. instructionless) basic blocks -- I believe this is the correct behavior? 

Also, it was previously possible to create a naming conflict in the CFG if there was a label `block0` and an unlabeled basic block. Prepending `label_` to the names of labeled basic blocks should fix that.

### DIFF
--- a/bril-ocaml/lib/func.ml
+++ b/bril-ocaml/lib/func.ml
@@ -14,7 +14,7 @@ type t = {
 let instrs { blocks; order; _ } = List.concat_map order ~f:(Map.find_exn blocks)
 
 let process_instrs instrs =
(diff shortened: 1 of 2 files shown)


sampsyo:
Preserving labels "blindly" (even when blocks are empty) seems like a good idea to me! You can imagine that a special CFG simplification pass could take care of removing empty BBs, taking care of cases where both labels appear to be necessary because they are attached to jumps/branches elsewhere in the function.

PR #89, opened 2020-11-27
Bril<->Rust extension

Pat-Lafon:
I figure if Bril has a Ocaml library, it could probably use a Rust library as well. This supports the main language extensions besides the recent speculation extension. 

I'm not quite sure what tests to add to show that core+extensions are all parse correctly though I've used it in the previous tasks and a bril interpreter so I'm reasonable confident in it.


sampsyo:
Awesome; thanks!!

PR #92, opened 2020-12-04
update bril-vim/syntax/bril.vim for speculate

mdmoeller:
Added keywords commit, guard, speculate to the vim syntax file.


sampsyo:
Wahoo; thanks!

PR #95, opened 2020-12-09
Bril rust extension

Pat-Lafon:
Adding the speculation extension. Introduces feature gating on each of the extensions so you only get what you need. Adding a serializing to stdout function seems like the bare minimum to include.


sampsyo:
Neato! I like the idea of putting the Bril language extensions behind Rust feature flags.

PR #98, opened 2020-12-20
brilirs: Remove custom parser and use bril-rs

### LINKED ISSUE #51: Fix the Rust interpreter for the new syntax
A "fun" task for any Rustaceans out there: #49 introduced a breaking syntax change, and the `brilirs` interpreter needs updating to match. In particular, the main necessary change is looking for labels in an instruction's `labels` key rather than in `args`.

yati-sagade:
This fixes #51.

Some ops such as "call" are unimplemented, but they shouldn't be too
difficult to add, looking at the TS reference interpreter code.

I copied test bril files into a separate dir here to keep tests scoped
to what we actually have implemented, although we likely need more
tests, e.g. for error-checking.


Pat-Lafon:
What does the ```Nil``` type mean in bril-rs?

yati-sagade:
@Pat-Lafon I added the `Nil` type for the `nil` value. Looking again though, I don't see any occurrences of `nil` in the TypeScript interpreter code, nor in the test/ dir. I think it might be an obsolete feature of the language, and probably can just be removed.

yati-sagade:
> What does the `Nil` type mean in bril-rs?

I am not sure if the language still all

brilirs/src/cfg.rs
```diff
@@ -1,39 +1,32 @@
...
+        last_instr
+      {
+        match op {
+          bril_rs::EffectOps::Jump | bril_rs::EffectOps::Branch => {
+            for l in labels {
+              block.exit.push(label_to_block_idx[l]);
+            }
           }
+          bril_rs::EffectOps::Return => {}
+          // TODO(yati): Do all effect ops end a BB?
```
sampsyo, line 25: No—for example, `print` is an effect operation but not a terminator.

brilirs/src/interp.rs
```diff
@@ -83,27 +63,291 @@ where
...
+    Return => {
+      out.flush().map_err(|e| InterpError::IoError(Box::new(e)))?;
+      // NOTE: This only works so long as `main` is the only function
+      return Ok(false);
+    }
+    Print => {
+      write!(
+        out,
+        // NOTE: The Bril spec implies print should just output its arguments, with no newline.
+        // However, brili uses console.log, which does add a newline, so we will too
```
sampsyo, line 328: I know you didn't write this comment, but FWIW, I recently clarified the docs to indicate that it should print a newline.

brilirs/src/interp.rs
```diff
@@ -114,152 +358,46 @@ pub fn execute<T: std::io::Write>(
...
-          out.flush().map_err(|e| InterpError::IoError(Box::new(e)))?;
-          // NOTE: This only works so long as `main` is the only function
-          return Ok(());
-        }
-        Id {
-          params: ValueOp { dest, typ, args },
-        } => {
-          if args.len() != 1 {
-            return Err(InterpError::BadNumArgs(1, args.len()));
+            // TODO(yati): Correct only when main is the only function.
```
sampsyo, line 395: Indeed! Totally cool to omit function calls for now.

sampsyo (commented):
Yay! This is fantastic! Thanks for tackling this. Wahoo!

Leaving out `call` seems right for now. I'll also open an issue about adding that someday.

PR #104, opened 2020-12-28
Everything but spec and benchmarks

Pat-Lafon:
I'm not quite sure how to split up a commit into readable pieces but I'll try to do so. I've mostly copied code from my previous attempt https://github.com/Pat-Lafon/bril/tree/brilirs/brili-rs. 

This adds the following:
- Allows brilirs to use the same test cases as brili using turnt 1.6.0
- A program is now composed of a hashmap of functions. Each function contains a list of its basic blocks and type information.(As opposed to a program containing one list of basic blocks for all of its functions).
- Previously labels were tossed out, they are now stored as an optional field of each basic block. 
- brilirs now accepts arguments from the command line for the main function(For a fully functioning test suite).
- brilirs errors now produce an error code of 2(previous 1).
- Function calls are now supported with corresponding errors.
- The SSA extension is now supported(phi instructions).
- The Memory extension is now supported with corresponding errors.
- Printing multiple values is now space separated and not comma separated.
- All `test/interp`, `test/mem`, and `test/fail` tests pass with the exception of the speculation extension tests.

Profiling(-p) for benchmarks is still unimplemented.
The Speculation extension is still unimplemented.
There basically been no consideration to performance.(Getting rid of all of the String::From() would be a huge upgrade).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ yarn.lock
 .pytest_cache/
 hypothesis_*
 __pycache__
+.vscode
--- a/benchmarks/turnt_brilirs.toml
+++ b/benchmarks/turnt_brilirs.toml
@@ -1,0 +1,3 @@
+command = "bril2json < {filename} | cargo run --manifest-path ../../brilirs/Cargo.toml -- -p {args}"
+output.out = "-"
+output.prof = "2"
--- a/bril-rs/src/lib.rs
+++ b/bril-rs/src/lib.rs
@@ -163,7 +163,7 @@ pub enum Literal {
 
 impl Literal {
     pub fn get_type(&self) -> Type {
-        match *self {
+        match self {
             Literal::Int(_) => Type::Int,
             Literal::Bool(_) => Type::Bool,
             #[cfg(feature = "float")]
--- a/brilirs/Cargo.toml
+++ b/brilirs/Cargo.toml
@@ -8,12 +8,6 @@ edition = "2018"
 
 [dependencies]
 clap         = "~2.31"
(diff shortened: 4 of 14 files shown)


sampsyo:
Yo! I think this is a lovely division of labor! All this stuff seems really great and we should merge it in. Apologies for the conflicts that arose in merging #103, though… there was a bit of duplicated work there, alas. Any chance you'd be interested in syncing things up?

It would also be awesome to update the docs (for both the library and the interpreter) to list the current set of supported extensions.

Miscellaneous loose ends, some of which probably deserve their own issues:

* We should probably reorganize the interpreter tests based on the extensions they use. That would make it easier to decide which directories need to be tested under both interpreters.
* Totally cool to leave off speculation, IMO. It's weird to implement, even in the TypeScript interpreter.
* For performance, good first steps would include (a) head-to-head comparison with the reference interpreter, and (b) simple function-level profiling.

Pat-Lafon:
I've rebased this against the current master branch. I've trimmed a couple things:
- We weren't really using the logging crates and the --verbose flag did very little so those have been removed.
- Previously a `ret` instruction was inserted at any exit block that did not have one. This will add to the dynamic instruction count without any effect so it's been removed.
- This `ret` instruction was needed because we returned an `EffectResult` enum which described when a function was ready to return. However, we have a control flow graph so we know when a function will return. This is enum has been replace with the Option type.
- `assert!`'s have been replaced with conditional checks to return an `InterpError` instead of an assertion error.

sampsyo (approved):
Fantastic! Thanks for getting this up to date!

Issue #99, opened 2020-12-21
LVN missed opportunity for eq?


Xazax-hun:
As far as I understand, if two values have the same value number, they are guaranteed to be the same at runtime. As a result, for operator `eq`, if both operands have the same value number (even if they are not constants) we could fold the operator to a constant true.

Xazax-hun:
Also for comparison operators, we might be able to fold when only one argument is constant. E.g. `false && var`.

sampsyo:
Yeah, that seems quite plausible! I'd be happy to merge those identities in the example LVN implementation if they're not too tricky to implement.

cgyurgyik:
Since `bril` doesn't support logical operators (please correct me if I'm wrong), this should be OK to close?

sampsyo:
We do have `and`, `or`, and `not`: https://capra.cs.cornell.edu/bril/lang/core.html#logic

cgyurgyik:
> https://capra.cs.cornell.edu/bril/lang/core.html#logic

Well that's embarrassing. I'll get a patch in for this.

PR #106, opened 2020-12-30
Add ne, eq, fix fold bug.

cgyurgyik:
`bril2json < comparisons.bril | python3 examples/lvn.py -p -c -f | bril2txt`

```
@main {
  a: int = const 4;
  b: int = const 3;
  c: bool = eq a b;
  print c;
}
```

=>

```
@main {
  a: int = const 4;
  b: int = const 3;
  c: bool = const false;
  print c;
}
```

Initially this was not folding since `const == False`.

### DIFF
--- a/examples/lvn.py
+++ b/examples/lvn.py
@@ -209,7 +209,8 @@ def _fold(num2const, value):
             return FOLDABLE_OPS[value.op](*const_args)
         except KeyError:  # At least one argument is not a constant.
             # Two values are equal if their arguments are equivalent.
-            if value.op == 'eq': return FOLDABLE_OPS[value.op](*value.args)
+            if value.op == 'eq' and value.args[0] == value.args[1]:
+                return True
             return None
         except ZeroDivisionError:  # If we hit a dynamic error, bail!
             return None
--- a/examples/lvn.py
+++ b/examples/lvn.py
@@ -15,6 +15,7 @@ class Numbering(dict):
     """A dict mapping anything to numbers that can generate new numbers
     for you when adding new values.
     """
+
     def __init__(self, init={}):
         super(Numbering, self).__init__(init)
         self._next_fresh = 0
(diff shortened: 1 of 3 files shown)


cgyurgyik:
Address #99 .

examples/lvn.py
```diff
@@ -205,6 +208,9 @@ def _fold(num2const, value):
             const_args = [num2const[n] for n in value.args]
             return FOLDABLE_OPS[value.op](*const_args)
         except KeyError:  # At least one argument is not a constant.
+            if value.op in {'eq', 'ne'} and value.args[0] == value.args[1]:
+                # Equivalent arguments may be evaluated for equality.
```
sampsyo, line 212: Maybe worth including an example, because this can be a little surprising? That is, what we're looking for is stuff like `eq x x` where `x` need not be a constant—it can be any variable. The weird thing is that this is not _exactly_ a value numbering thing (it depends on variables, not values), although it is convenient to do here.

sampsyo (commented):
Seems great! What do you think about adding your test cases from the issue body as test files herein?

examples/lvn.py
cgyurgyik: Done. Wouldn't this fall under constant folding rather than LVN?

cgyurgyik:
> Seems great! What do you think about adding your test cases from the issue body as test files herein?

Is there some place specifically LVN is tested?

sampsyo:
Ah yeah, good point: I have them here. https://github.com/sampsyo/bril/tree/master/examples/test/lvn

cgyurgyik:
> Ah yeah, good point: I have them here. https://github.com/sampsyo/bril/tree/master/examples/test/lvn

Thanks! Didn't see those. Done.

examples/lvn.py
sampsyo: Yeah, it's a little fuzzy. It's just that all the other constant folding we do *during* LVN is based on value numbers, not on variable identity. So this one breaks the mold.

sampsyo (approved):
Wahoo; perfect! Thanks!!

PR #107, opened 2021-01-02
Algebraic types extension

Calsign:
Work from my [final project](https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/bril-algebraic-data-types/).

Adds the following algebraic type extension features:

 - Support in `brili` and `bril-txt`
 - Support in the JS, OCaml, and Rust libraries
 - Type-checker `briltc`
 - Rust compiler `rs2bril`
 - Bril samples
 - Rust samples

The standalone tools are currently in the `algebraic-types` directory, but I can reorganize things if desired.


algebraic-types/README.md
```diff
@@ -0,0 +1,7 @@
+
+Extra tools and samples for the Bril algebraic types extension.
+
+ - `briltc`: Bril type-checker with support for algebraic types
```
sampsyo, line 4: Seems like this can just be at the top level? We do have `type-infer` there, and this is similar to that but different. The fact that it supports the ADT extension is not terribly important; someday, people might extend other stuff with similar support (or extend this with support for other extensions).

algebraic-types/README.md
```diff
@@ -0,0 +1,7 @@
...
+ - `rs2bril`: compiler from Rust syntax to Bril
+ - `samples`: sample Bril programs using algebraic types
+ - `samples-rs`: sample Rust programs
```
sampsyo, line 7: Those can be thought of as tests for `rs2bril`, right?

algebraic-types/README.md
```diff
@@ -0,0 +1,7 @@
+
+Extra tools and samples for the Bril algebraic types extension.
+
+ - `briltc`: Bril type-checker with support for algebraic types
+ - `rs2bril`: compiler from Rust syntax to Bril
+ - `samples`: sample Bril programs using algebraic types
```
sampsyo, line 6: Seems like maybe these should be included in the test suite? Or are they different?

algebraic-types/README.md
```diff
@@ -0,0 +1,7 @@
+
+Extra tools and samples for the Bril algebraic types extension.
+
+ - `briltc`: Bril type-checker with support for algebraic types
+ - `rs2bril`: compiler from Rust syntax to Bril
```
sampsyo, line 5: Seems like this deserves to be at the top level too. It's a sibling to `ts2bril`.

algebraic-types/rs2bril/Makefile
```diff
@@ -0,0 +1,14 @@
+
```
sampsyo, line 1: I know Makefiles are cool, but these are kinda the standard commands one would run for any Cargo-based Rust project. So maybe not necessary?

algebraic-types/README.md
```diff
@@ -0,0 +1,7 @@
+
```
sampsyo, line 1: I would love to integrate all these tools in with the rest of the tools here, rather than keeping them in a separate ADT-specific sandbox. Justifications below.

bril-ts/brili.ts
```diff
@@ -708,12 +749,84 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
...
+
+  case "construct": {
+    if (!instr.args || instr.args.length < 2) {
+      throw error("malformed construct instr");
+    }
+    if (typeof(instr.type) !== "object" || !instr.type.hasOwnProperty("sum")) {
+      throw error("construct destination must be a sum type");
+    }
+    let value = getArgument(instr, state.env, 0);
+    let index = parseInt(instr.args[1]);
```
sampsyo, line 786: TBH, I am still a little concerned about this syntax change. I don't have an alternative to propose, unfortunately, but it is pretty weird that _for every other instruction_, arguments are variable names, but for _this one special instruction_, the argument is a string that encodes a decimal integer and not a variable name at all. For example, consider writing a def/use analysis: you would now have to say "all the things in `args` are variable uses, *except for this one instruction case*, where one of the arguments does *not* name a variable."

Like I said, not sure what to do about this, but I think we may want to reconsider the abstract syntax for this instruction…

bril-txt/briltxt.py
```diff
@@ -29,16 +30,17 @@
 eop.2: op ";"
 label.1: LABEL ":"
 
-op: IDENT (FUNC | LABEL | IDENT)*
+op: IDENT (FUNC | LABEL | IDENT | SIGNED_INT)*
```
sampsyo, line 33: This is the concrete-syntax reflection of the abstract-syntax complaint I have above.

bril-txt/briltxt.py
```diff
@@ -29,16 +30,17 @@
...
 ?tyann: ":" type
 
 lit: SIGNED_INT  -> int
   | BOOL         -> bool
   | SIGNED_FLOAT -> float
 
-type: IDENT "<" type ">" -> paramtype
-    | IDENT              -> primtype
+type: IDENT "<" type ("," type)* ">" -> paramtype
+    | IDENT "<" ">"                  -> paramtype
```
sampsyo, line 42: Seems legit, but maybe this grammar could be a little simpler to allow "zero or more" types? (As in other locations, such as function arguments, where we allow zero or more of things.)

bril-txt/briltxt.py
```diff
@@ -146,7 +148,10 @@ def bool(self, items):
             return False
 
     def paramtype(self, items):
-        return {items[0]: items[1]}
+        if len(items) == 2:
+            return {items[0]: items[1]}
```
sampsyo, line 152: Ah, this is a little worrisome. It means it's not possible to have one-argument parameterized types. Like, doesn't `product<int>` now do something wrong (even if that's not a very practical type)? Not sure what to do about this, TBH… maybe the whole parameterized-type thing needs rethinking. 🙃

sampsyo (commented):
Super cool; thanks for the PR! I do think it would make sense to reorganize things so they get integrated with the rest of the tools… for most of this stuff, there's not much ADT-specific about them!

There are two design things that I am currently worried about, noted within:

* The abstract syntax for `construct` overloads `args` to contain not just variable names but also string-encoded integers. This is a one-off exception to a rule that otherwise serves us pretty well. TBH I am not sure what to do about this, because you do want to have a static integer as a parameter there. This requires some creativity. 😕 Maybe we want instructions to optionally have a separate list of static parameters for cases like this.
* The concrete syntax for parameterized types is a little sad because there is a special case for one-argument types vs. everything else. This isn't too bad though because it's just about the text format, not the AST.

One other thing that will need to happen before merging is adding to the documentation. We will need book pages for the language extension and each of the new tools.

Issue #109, opened 2021-01-03
Domination frontier including self in frontier.


cgyurgyik:
Hi! Checking my results against the `examples` again. 

From the lecture notes,
We say `A`'s domination frontier contains `B` if 
1. `A` **does not** strictly dominate* `B`
2. `A` dominates a predecessor of `B`.

*I think this was just `dominates` rather than `strictly dominates` in the video lecture. However it is defined correctly in the [Gist](https://www.cs.cornell.edu/courses/cs6120/2020fa/lesson/5/)

```bril2json < class-example.bril | python3 examples/dom.py front```

```
# Example used for Global Analysis in L5.
@main {
.entry:
  x: int = const 0;
  i: int = const 0;
  one: int = const 1;
  jmp .loop;

.loop:
  max: int = const 10;
  cond: bool = lt i max;
  br cond .body .exit;

.body:
  mid: int = const 5;
  cond: bool = lt i mid;
  br cond .then .endif;

.then:
  x: int = add x one;
  jmp .endif;

.endif:
  factor: int = const 2;
  x: int = mul x factor;
  i: int = add i one;
  jmp .loop;

.exit:
  print x;
  ret;
}
```

Results in:

```
# Frontiers
{
  "body": [
    "loop"
  ],
  "endif": [
    "loop"
  ],
  "entry": [],
  "exit": [],
  "loop": [
    "loop" # ???
  ],
  "then": [
    "endif"
  ]
}

```

Does `B = loop` meet the frontier criteria for `A = loop`?

1. `loop` **does not** strictly dominate `loop` <-- does hold since, `loop == loop`.
2. `loop` dominates an immediate predecessor of `loop`. <-- does **not** hold. The only predecessor to `loop` is `entry`, which is not dominated.

Let me know if something looks off with my explanation.

cgyurgyik:
After some thinking, I take back what I said above about a node not being able in its own dominance frontier. That should be possible. 

![image](https://user-images.githubusercontent.com/37983775/103486630-32bfd800-4dcd-11eb-9dc3-2db8d4c81f1e.png)


Let `A = entry`, `B = entry`.
1. `entry` does not strictly dominate `entry` <-- does hold, since `entry == entry`
2. `entry` dominates an immediate predecessor of `entry`. <-- does hold, since `entry` dominates `loop`.

Therefore, `entry` is in its own dominance frontier.
That doesn't hold for the original example above.

sampsyo:
[...]

PR #110, opened 2021-01-03
Fix domination frontier bug.

cgyurgyik:
Discussed in #109.

`frontiers[block] = [b for b in dominated_succs if b not in dom_inv[block]]`

To my understanding, `dom_inv[b]` should contain `b` itself since `A dominates A` for all `b`. This avoids adding itself to the frontiers of `b`.

I took a different approach to this problem, so I may not be following your algorithm correctly.


cgyurgyik:
Incorrect.

Issue #108, opened 2021-01-03
[SSA] Buggy SSA output with `to_ssa.py` and a modified `if-ssa.bril`


terrelln:
This input, which is exactly [if-ssa.bril](https://github.com/sampsyo/bril/blob/d65455d1eb9c4b78711af2fffc7453bbf205775e/examples/test/ssa/if-ssa.bril) except the `.exit` label is renamed to `.xit` causes buggy SSA to be generated by `to_ssa.py`:

```
@main(cond: bool) {
.entry:
    a.1: int = const 47;
    br cond .left .right;
.left:
    a.2: int = add a.1 a.1;
    jmp .xit;
.right:
    a.3: int = mul a.1 a.1;
    jmp .xit;
.xit:
    a.4: int = phi .left a.2 .right a.3;
    print a.4;
}
```

The generated SSA is:
```
@main(cond: bool) {
.entry:
  a.1.0: int = const 47;
  br cond .left .right;
.left:
  a.2.0: int = add a.1.0 a.1.0;
  jmp .xit;
.right:
  a.3.0: int = mul a.1.0 a.1.0;
  jmp .xit;
.xit:
  a.2.1: int = phi a.2.0 a.2.0 .left .right;
  a.4.0: int = phi a.2.1 a.3.1 .left .right;
  print a.4.0;
  ret;
}
```

The buggy line is `a.4.0: int = phi a.2.1 a.3.1 .left .right;`. When called with `cond = false` the variable `a.3.1` is not defined, causing `a.4.0` to be undefined.

This is caused by the function `prune_phis()`:

https://github.com/sampsyo/bril/blob/d65455d1eb9c4b78711af2fffc7453bbf205775e/examples/to_ssa.py#L91-L126

This bug is triggered with the new input, and not triggered by `if-ssa.bril` because of the sorting done on this line:

https://github.com/sampsyo/bril/blob/d65455d1eb9c4b78711af2fffc7453bbf205775e/examples/to_ssa.py#L79

I believe that this optimization is not sound, even in the case when the input does not have `phi` instructions. Imagine the input:

```
@main() {
    cond: bool = const true;
    br cond .true .false;
.true:
    a: int = const 0;
    jmp .zexit;
.false:
    b: int = const 1;
    jmp .zexit;
.zexit:
    print a;
}
```

Because `cond` is always `true`, `a` is always defined. But, `to_ssa.py` generates incorrect code similar to before:

```
@main {
.b1:
  cond.0: bool = const true;
[...]

PR #111, opened 2021-01-03
[SSA] Fix bug in to_ssa.py

terrelln:
Fixes Issue #108.

The optimization in prune_phis() is not sound. phi nodes that are
"partially undefined" cannot be removed. It is only illegal to read
from the result if the second to last executed label corresponds to
an undefined argument. `to_ssa.py` generated incorrect code for the
two tests added, before the fix.

In most cases, these phi nodes aren't read from, and will be removed
by DCE. However, in the case where the phi nodes are read, a correct
optimization would be:
1. Detect partially undefined phi nodes whose value is always used.
   Let `undefined_labels` be the set of labels whose argument is
   undefined.
2. Leverage the undefined behavior to say that `undefined_labels`
   are in fact not predecessors to the basic block.
   `preds = preds - undefined_labels`.

Though, I'm not sure how useful that optimization would be in
practice.

### DIFF
--- a/examples/to_ssa.py
+++ b/examples/to_ssa.py
@@ -74,6 +74,9 @@ def ssa_rename(blocks, phis, succ, domtree, args):
             for p in phis[s]:
                 if stack[p]:
                     phi_args[s][p].append((block, stack[p][0]))
+                else:
+                    # The variable is not defined on this path
+                    phi_args[s][p].append((block, "__undefined"))
 
         # Recursive calls.
         for b in sorted(domtree[block]):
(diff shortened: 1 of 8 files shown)


terrelln:
You could also find phi nodes that have only one non-undefined argument and replace it with an `id` instruction.

terrelln:
> You could also find phi nodes that have only one non-undefined argument and replace it with an id instruction.

You'd need to merge PR #112 to allow that optimization. You can decide if you think it is a good idea or not.

examples/to_ssa.py
```diff
@@ -74,6 +74,9 @@ def _rename(block):
             for p in phis[s]:
                 if stack[p]:
                     phi_args[s][p].append((block, stack[p][0]))
+                else:
+                    # The variable is not defined on this path
+                    phi_args[s][p].append((block, "__undefined"))
```
terrelln, line 79: This isn't 100% necessary, as `brili` supports `phi` nodes with incomplete sets of labels. If the label isn't present it treats the variable as undefined. However, the bril spec says that "It is an error to use a phi instruction when [..] the instruction does not contain an entry for the second-most-recently-executed label".

Let me know what you would prefer.

sampsyo: This is perfect! I can definitely see it going either way, but I like that this version of the pass, at least, maintains the invariant that phi-nodes reflect the in-degree of the CFG node they are attached to (as they do in LLVM, for example).

sampsyo (approved):
Awesome; thank you so much for sorting this out!! Really nice work.

PR #112, opened 2021-01-03
[bril-ts][brili] Allow unused id of undefined

terrelln:
Allow `id` instructions with undefined arguments, similar to `phi`
instructions. This allows optimizers to translate `phi` instructions
with only one non-undefined argument into `id` instructions. It is
an error to use the destination variable when it is undefined.

I've added successfull and error test cases for both phi and id nodes
with undefined values.

An example use case is if the code is already in SSA form. `to_ssa.py`
might generate something like this:

```
a.1: int = phi a.0 __undefined .true .false;
b.1: int = phi __undefined b.0 .true .false;
c.1: int = phi a.1 b.1 .true .false;
```

Instead it could be turned into:

```
a.1: int = id a.0;
b.1: int = id b.0;
c.1: int = phi a.1 b.1 .true .false;
```

It isn't strictly necessary for brili to support this, but I think it is
a logical feature to support, since `phi` nodes already support it.


sampsyo:
However, to set an intention here: we should probably do _something_ at some point to clean up those `id`s of undefined values. To make it explicit, this is the other side of the coin and the downside to making `id` and `phi` behave differently: running "from-SSA" converts the latter into the former, turning benign behavior into an error.

A dumb way to do this would be to have the "from-SSA" pass avoid creating `id` copies of any variable that is globally undefined (because this would be an error anyway).

terrelln:
> The reason why I'm comfortable making an exception for phi is because interpreting SSA-form programs is a pretty weird thing to do!

That makes sense to me.

> A dumb way to do this would be to have the "from-SSA" pass avoid creating id copies of any variable that is globally undefined (because this would be an error anyway).

The SSA algorithm also creates lines that reference undefined variables that aren't globally undefined (depending on the order the blocks are processed). For example in this test case:

https://github.com/sampsyo/bril/blob/5d4c9cf3a0576c1f28b6caf4d0e72faf100ca957/examples/test/to_ssa/if-const.out#L1-L16

`b.1` is assigned from `b.0` from both predecessors, but is only defined in one of them. It seems like a naive from-SSA algorithm would generate broken code in this case as well. Would an uninitialized variables pass on only these id nodes be enough? I will think about it when I implement my from-SSA function.

sampsyo:
Ah yes, great point. That more completely reveals the mismatch between the undefined-tolerant `phi` and an undefined-rejecting `id`. Requires even more thought! Thanks for bringing up this interesting issue.

terrelln:
> Ah yes, great point. That more completely reveals the mismatch between the undefined-tolerant phi and an undefined-rejecting id. Requires even more thought! Thanks for bringing up this interesting issue.

In my [from SSA function](https://github.com/terrelln/cs6120/blob/efee1d1aaef9fe0a3891551332cb12705f0a925f/src/ssa.rs#L256-L287) I ended up just doing an "initialized variables" dataflow analysis, that tells me which variables may be initialized after each block. And I'll only create `id` nodes for variables which may be initialized (or were just defined by a previous `id` node in the same block).

I don't think that this solves the problem in 100% of cases, since the may be initialized analysis is conservative, but it isn't trivially broken, and it works for my test cases.

Issue #105, opened 2020-12-30
[LVN] Missed constant fold


cgyurgyik:
I was checking my implementation against the examples, and found this:
`bril2json < overwritten-variable.bril | python3 examples/lvn.py -p -c -f | bril2txt`

```
@main {
  v1: int = const 4;
  v2: int = const 0;
  mul1: int = mul v1 v2;
  v2: int = const 3;
}
```

Lowers to:
```
@main {
  v1: int = const 4;
  lvn.2: int = const 0;
  mul1: int = mul v1 lvn.2; // <-- This should be folded.
  v2: int = const 3;
}
```

Expected output: 
```
@main {
  v1: int = const 4;
  lvn.2: int = const 0;
  mul1: int = const 0;
  v2: int = const 3;
}
```

cgyurgyik:
Let me know if I'm wrong here. I'll come back to this later if so, I need a break from LVN for a little :-)

sampsyo:
Huh, yeah, it does look like it should do that. I remember wrestling with something similar earlier, perhaps in the context of #77, but I don't really know why it's missing this.

PR #114, opened 2021-01-05
Test for #105.

cgyurgyik:
#105 was fixed with #106. Here's an accompanied test.
Since `mul1 == 0`, the conditional `if const: ...` was false.


sampsyo (approved):
Wonderful; thanks!

Issue #77, opened 2020-09-16
Fix example LVN pass's copy propagation on block inputs


sampsyo:
In the examples, `lvn.py -p` currently fails for this example:

```
@main {
  a: int = const 42;
.lbl:
  b: int = id a;
  a: int = const 5;
  print b;
}
```

The optimization incorrectly changes the last line to `print a`.

The problem is the interaction between copy (`id`) propagation and basic block inputs (i.e., the fact that `a` is read before it is written in the second block). One annoying but cleanish solution would be to copy `a` to a fresh variable right as the block begins because it is read before it is written. Another, much less clean solution would be to detect the first overwrite of `a` and do the renaming at that point, or to somehow avoid doing copy propagation for the `id` because the value will be clobbered later.

PR #116, opened 2021-01-06
Attempt at fixing copy propagation in LVN pass.

cgyurgyik:
This is a first attempt at #77. It is far from elegant.
This gist of the function is to:

1. Look for an instruction with `id` op. If `arg` has not been defined yet, assume it is defined outside of the local scope.
2. For every other instruction, if the `dest` name is equivalent to some variable defined outside the local scope, rename it.
3. Rename every use until its next assign as well.

### DIFF
--- a/examples/lvn.py
+++ b/examples/lvn.py
@@ -64,6 +64,56 @@ def read_first(instrs):
     return read
 
 
+def rename_nonlocal_duplicates(block):
+    """Renames any variables that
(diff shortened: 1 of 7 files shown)


sampsyo:
Cool. As a warning, giving this its full consideration will require some deep thinking that I may not be able to get to for a while. So please don't fret if I don't get around to merging this soon.

cgyurgyik:
No problem!

Issue #91, opened 2020-11-27
Bril Type Error questions


Pat-Lafon:
2 benchmarks (perfect.bril and digital-root.bril) have ret instructions at the end of their main functions that return values despite them not having a return type. I guess this is a type error?

A more general question. From the type system perspective, if a function both returns a value and has a side effect, does it need to be called as a Value operation (because it returns something) or can it be called as an Effect operation (and just ignore the return value)?

sampsyo:
1. Yeah, that's a good point—functions should not return a value if they don't have a return type declared. It's worth re-emphasizing that both of those benchmarks you mentioned have the weird `ret` in the `main` function, which isn't supposed to return anything. Maybe someday it would make sense to allow returning a value from main, but if we do that, then it would require declaring the return type before doing that.
2. The spec should be clearer about this, but I think the intent is that `call` *must* be used as a Value Operation if the function returns a value, i.e., return values are not ignorable.

Pat-Lafon:
As I'm running the benchmarks with brilirs, I've noticed it breaks on lines like the following `v21: float = const 2;` because this looks like an integer literal being promoted to a float. Is the previous line allowed or should only `v21: float = const 2.0;` be valid?

sampsyo:
Right, I noticed that too when fiddling around with it. I think the Right Thing to Do is to make that illegal, i.e., float constants must be floats like 42.0 and integer constants must be integers like 42 (maybe the docs need amending). Checking this is annoyingly hard in TypeScript because of JavaScript's all-float-all-the-time `number` type, but of course it's trivial with `serde_json`. So IMO let's leave it be.

Pat-Lafon:
With #117 merged I'm closing this issue.

PR #117, opened 2021-01-06
brilirs + benchmarks

Pat-Lafon:
This pr allows `brilirs` to be run against the same benchmarks as `brili` and contains the following changes:
- Relating to #91, a couple of benchmarks which called `ret` at the end of their `main` function have had this instruction removed. 
- ~Relating to a follow up question in #91, a few benchmarks now consistently have a decimal to designate their floats as floating point literals. If `v21: float = const 2;` is supposed to be allowed, this can be reversed and brilirs updated.~ `brilirs` now allows integer literals to be promoted to floating point values.
- One benchmark, `mat-mul.bril` had a comma in one of it's arguments to main and has been removed. This didn't affect `brili` because parseInt() will truncate off any additional characters that it can't parse.
- The `benchmark/` suite can be run with `make benchmark`. (You need to pass the `--quiet` flag to cargo for stderr output to work with turnt)
- `brilirs` now accepts the `-p` flag to output the total number of dynamic instructions.
- Edge case bugs in cfg creation have been fixed.
- There was a bug with the type of floating point logical instructions.

`brilirs` can now run correctly run all but ~2 benchmarks~ 1 benchmark:

~`quadratic.bril` passes a negative argument which the clap cli seems to have issues accepting. `allow_hyphen_values` should allow this to work but I haven't managed to debug why it doesn't.~ `brilirs` now takes negative values from the cli as expected.

`mat-mul.bril` does a matrix multiplication which overflows as an i64 in rust. As https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/floats-static-arrays/ mentions, the BigInt type on it's own will not correctly handle `bril`'s int specification of "64-bit, two’s complement, signed integers." Bril floats probably have a similar issue since they are implemented with number which I believe is 53 bit. Therefore, I believe the output of this benchmark to be incorrect.

### DIFF
--- a/benchmarks/digital-root.bril
+++ b/benchmarks/digital-root.bril
@@ -30,7 +30,6 @@
 
  .done:
     print result;
-    ret result;
 }
 
 @is_single_digit(input: int): bool {
--- a/benchmarks/digital-root.prof
+++ b/benchmarks/digital-root.prof
@@ -1,1 +1,1 @@
-total_dyn_inst: 248
+total_dyn_inst: 247
--- a/benchmarks/mat-mul.bril
+++ b/benchmarks/mat-mul.bril
@@ -54,7 +54,7 @@
   jmp .loop;
 .done:
   ret;
-} 
+}
 
 # Naive matrix muliplication algorithm
 # `arr1`, `arr2`, and `dest` are `size` x `size` matrices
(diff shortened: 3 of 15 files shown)


Pat-Lafon:
With the cli change to brilirs, I don't think there are any remaining issues with this pr? I don't expect there to be a resolution with the `mat-mul` benchmark(unless I'm mistaken on the issue).

sampsyo:
Yay! Thanks for all these additions. And you're right; the `mat-mul` benchmark itself needs to be fixed rather than this interpreter. :smiley:

PR #118, opened 2021-01-11
Address latter part of #108.

cgyurgyik:
Originally, values were added to `phi` that were undefined along the path.

Originally I was trying to force the use of reaching definitions, but this shouldn't be necessary. All we care about here is that for each `p`, it was defined along some path beforehand. Otherwise, it is, well, undefined. An example:

```
@main() {
    cond: bool = const true;
    br cond .true .false;
.true:
    a: int = const 0;
    jmp .zexit;
.false:
    # `a` is undefined along this path.
    jmp .zexit;
.zexit:
    print a;
}

```
Outputs
```
@main {
.b1:
  cond.0: bool = const true;
  br cond.0 .true .false;
.true:
  a.0: int = const 0;
  jmp .zexit;
.false:
  jmp .zexit;
.zexit:
  a.1: int = phi __undefined a.0 .false .true; # Before, this was `phi a.0 a.0 .false .true`
  print a.1;
  ret;
}
```


cgyurgyik:
I've tried quite a few benchmarks in SSA form with two additional changes not in this PR:
1. Don't append the argument if it is `__undefined`.
2. If # pairs in `phi`'s arguments is less than or equal to one, don't even add `phi`.

This makes me think that these additional changes are OK as well. However, when I then convert out of SSA using `python3 from_ssa.py`, I begin to get undefined variables in the benchmarks. I haven't pinpointed why this is occurring yet.

terrelln:
Isn't your `defined_analysis()` an over-estimate of the defined variables, so it will miss some cases?

I believe the problem you're trying to solve:

```
a.1: int = phi a.0 a.0 .true .false
```

Is caused by a bug in the `to_ssa.py` script. I'm going to put up a PR to fix it shortly.

terrelln:
I've put up PR #119. I believe that it fixes all the issues of #108.

cgyurgyik:
> Isn't your `defined_analysis()` an over-estimate of the defined variables, so it will miss some cases?
> 
> I believe the problem you're trying to solve:
> 
> ```
> a.1: int = phi a.0 a.0 .true .false
> ```
> 
> Is caused by a bug in the `to_ssa.py` script. I'm going to put up a PR to fix it shortly.

I'm not saying I'm correct in any way, but can you explain what you mean by "over-estimate of defined variables"? I'm not sure I'm following. Any yep! That's what I was attempting to fix.

terrelln:
> can you explain what you mean by "over-estimate of defined variables"? 

Consider the variable `v` in the `loop` function in this example. I believe your analysis will say that `v` is defined at the entry of every block except `.entry`. But, when `print` is false `v` is never defined (and isn't defined in the first iteration of the loop). In that case, it over-estimates the set of defined variables.

```
@func(): int {
    n: int = const 5;
    ret n;
}

@loop(infinite: bool, print: bool) {
.entry:
.loop.header:
    br infinite .loop.body .loop.end;
.loop.body:
    br print .loop.print .loop.next;
.loop.print:
    v: int = call @func;
    print v;
.loop.next:
    jmp .loop.header;
.loop.end:
}

@main() {
  infinite: bool = const false;
  print: bool = const true;
  call @loop infinite print;
}
```

I think a better name for the analysis is `maybe_defined_variables`. You can use it as a negation. E.g. any variable not in the set is certainly not defined. And conversely you can use the `maybe_undefined_variables` analysis to say anything not in that set is certainly defined.

cgyurgyik:
Huh, yeah good point. It produces the same SSA as your PR, but `defined` is a poor choice of terminology. Thanks for the insight.

PR #119, opened 2021-01-12
[SSA] Fix up `to_ssa.py` and add test cases

terrelln:
This PR contains 3 commits detailed below. In summary the goal is to remove all uses of undefined variables. It fixes all the problems of Issue #108. It is done in 2 steps:

1. Fix a bug in `rename()` that caused references to undefined variables (not `__undefined`) to be emitted.
2. Define `__undefined.${type}` for each type at the top of the entry block, and use `__undefined.${type}` instead of `__undefined`.

After these steps `phi` nodes no longer have undefined arguments, and no longer produce undefined results. The ability to execute `phi` nodes with undefined or missing arguments could be removed from `brili`, or hidden behind a flag (but I haven't done that).

I added one test case `to_ssa/` called `loop-branch.bril` which has an interesting feature where two phis form a dependency loop: `v.0` depends on `v.1` and visa-versa.

Then I add test cases for `from_ssa.py` that executes the round-tripped code for every test in `to_ssa/`. I also tested on all the programs in the `benchmarks` directory, and it works on everything in that directory, except for programs which require the `memory` extension, because as far as the docs say, there isn't a constant pointer value `null`. I haven't added them to the `from_ssa` tests to avoid adding overly complex tests.

This newly generated code inserts potentially unnecessary `__undefined.${type}` definitions. A DCE pass can trivially remove the unnecessary ones.

---
**[to_ssa] Fix bug in rename()**

The use of `defaultdict()` and `stack.update(old_stack)` meant that any
stacks that were empty at the top of `_rename()` weren't cleared at the
bottom. This caused use of undefined variables inserted into `phi`
instead of `__undefined`.

---
**[to_ssa] Define undefined variables**

Break undefined variables up by type, to meet the bril spec that each
variable has only one type. Then, define the undefined variable for each
type in the entry block. This ensures that no phi node uses undefined
variables, or produces undefined variables, in any case.

Add an interesting test case `loop-branch.bril` which produces phi nodes
that form a dependency loop.

---
**[from_ssa] Add test cases**

Add `brili` test cases, since I want to test that no undefined varaibles
are used during execution after `from_ssa`.


terrelln:
I think the allowance of undefined variables in `phi` nodes may have been a mistake. The wrinkle that `phi` nodes are allowed undefined values, but no other instruction is, really complicates analysis of the `bril` language.

I think that there are two logical choices, and anything in the middle will cause headaches:
1. Strictly allow no undefined variables.
2. Allow undefined variables in all ops. Any operation with undefined variables is undefined.

cgyurgyik:
Cool! This is probably much cleaner than my approach. The end result is the same, having to define an `__undefine`. Couldn't this be removed with a pass before emitting code in SSA form (as you mentioned earlier)?

examples/to_ssa.py
```diff
@@ -105,6 +106,28 @@ def insert_phis(blocks, phi_args, phi_dests, types):
...
+        "int": 0,
+        "bool": False,
+        "float": 0.0,
+    }
+    return values[var_type]
+
+
+def insert_undefined_vars(blocks, types):
+    entry = list(blocks.keys())[0]
+    types = sorted(set(types.values()))
```
cgyurgyik, line 120: I believe this will crash on some of the benchmarks that use pointers. For example,
https://github.com/sampsyo/bril/blob/dbbd8e8f1d7000fefdcffaef2d79ce3339e28b40/benchmarks/eight-queens.bril#L6
` bril2json < ../../../benchmarks/eight-queens.bril | python ../../to_ssa.py`
The pointer type is stored in a `dict`, and thus unhashable.

terrelln: Yes, is this something that `to_ssa.py` should support? If so I can add it. There currently isn't a `null` value for pointers, but I suppose `alloc 0;` would work, and just never `free`.

terrelln:
I've added support for the memory extension to `to_ssa.py`. In order to do that, I needed to support `null` pointers in `brili` and `brilirs`. Null pointers are created with `alloc 0`, and do not need to be freed. See the first commit message for details.

I've decided to use `alloc 0`, so as not to introduce another instruction, or make `const` memory aware.

sampsyo:
I know this issue is old, but I want to write up a summary of where things stand on this sad state of affairs. Basically, writing an SSA conversion in a way that deals elegantly with undefined values turns out to be pretty frustrating. To re-illustrate the basic problem, consider this simple program (called `if-const.bril` in our tests):

```
@main() {
    cond: bool = const true;
    br cond .true .false;
.true:
    a: int = const 0;
    jmp .exit;
.false:
    b: int = const 1;
    jmp .exit;
.exit:
    print a;
}
```

The thing to notice about this program is that it is only safe to run because the branch happens to always go to the `.true` side. So `a` is always defined in `print a`, even though it wouldn't be if the branch were to go the other way (which, again, is impossible).

If we follow our SSA conversion algorithm to the letter, we'd end up first renaming the assignment to `a` thus:

```
a.0: int = const 0;
```

And generating a phi-node that must look something like this in `.exit`:

```
a.1: int = phi .true a.0 .false ???;
```

But what should the value of `a.1` be if we traverse the edge from the `.false` block? The original variable `a` was undefined, so we need to simulate that!

Our current "solution" is to just replace that `???` with a reference to an *actually undefined variable*. So we hallucinate the name `__undefined` and generate a phi-node that looks like this:

```
a.1: int = phi .true a.0 .false __undefined;
```

…the idea being that this edge *will never actually be traversed*, so it should be fine to refer to a variable that doesn't actually exist here. Easy peasy, right?

Not so fast—this works until we try to generate the phi-node for `b` in `.exit`:

```
b.1: int = phi .true __undefined .false b.0;
```

Oh no! The CFG edge from `.true` *will* actually be traversed… so this SSA-ified program will end up referring to an undefined variable and crashing, even though the original program was just fine.

To make this bad scenario more concrete (because running programs directly in SSA form is kinda weird), consider converting this program *back out of SSA form*, which means inserting assignments into the predecessor blocks. Namely, here, we'll put this into the end of the `.true` block (remember, that's a block that actually runs):

```
b.1: int = id __undefined;
```

The interpreter will, understandably, not like this line.

Again, we have a clever "solution": let's just run dead code elimination! This `b.1` variable can't possibly be used. One way to put it is that it is *guaranteed* that this version of the original `b` variable was never used downstream of the `.true` block because, if it were, that original program would have been illegal by virtue of reading an undefined variable itself. So, we can just run our trivial dead code elimination pass to get rid of this pesky undefined variable! And that mostly works.

The only problem is that trivial dead code elimination is trivial. It can't identify more sophisticated situations where these needless generated assignments are inserted. The most dastardly case is where a *cycle* exists between two generated SSA variables, both of which are dead, but neither of which looks trivially dead in isolation. This comes up if you write a loop that assigns to a variable (that isn't assigned outside of the loop), like in this code from `loop-branch.bril`:

```
@loop(infinite: bool, print: bool) {
.entry:
.loop.header:
    br infinite .loop.body .loop.end;
.loop.body:
    br print .loop.print .loop.next;
.loop.print:
    v: int = call @func;
    print v;
.loop.next:
    jmp .loop.header;
.loop.end:
}
```

In this code, `v` needs a phi-node at the top of the loop (the `.loop.header` block). The value for `v` there needs to come either from the loop itself or from the function entry, in which case it's undefined. And `.loop.next` needs a phi-node of its own to pick between the `call` and the previous iteration of the loop, i.e., the phi-node at the top of the loop. Here's our current version of the SSA-ified code:

```
@loop(infinite: bool, print: bool) {
.entry:
  jmp .loop.header;
.loop.header:
  v.0: int = phi __undefined v.1 .entry .loop.next;
  br infinite .loop.body .loop.end;
.loop.body:
  br print .loop.print .loop.next;
.loop.print:
  v.2: int = call @func;
  print v.2;
  jmp .loop.next;
.loop.next:
  v.1: int = phi v.0 v.2 .loop.body .loop.print;
  jmp .loop.header;
.loop.end:
  ret;
}
```

The problem here is that, if you insert all the copies to reflect these phi-nodes, the `__undefined` copy will not be dead. `v.0` is used later in `v.1`, and then `v.1` is used in `v.0`. We have a dead cycle: all the code is dead, but it's "hard" to eliminate the code because it would require detecting cycles (rather than the simple "never used anywhere" check we do for trivial DCE). Annoying!

In #119, @terrelln had the bright idea to just define all these variables. SSA conversion would just insert actual variables like `__undefined.int` at the top of the function to give zero/null values that could be used by all these undefined cases. That way, we wouldn't be doing the weird hack where we refer to *literally* undefined variables ever, so code would *always* be safe from crashing after SSA conversion.

I like this solution well enough, but it has a couple of drawbacks:

* It requires a null value to exist for every type! As new types are added, the SSA code has to know how to generate null values for every type. This seems sort of inconvenient to maintain in the face of new extensions. In the memory extension in particular, it implies that we needed to add null pointers everywhere (the "billion dollar mistake").
* It seems vaguely unfortunate that we need to generate all this code for null values *that is guaranteed to be dead*. It seems like a lot of extra effort to create values that are certain never to be read by well-formed programs. For the sake of clarity of the conversion, it would be super great if we didn't have to intentionally generate all this dead code—if we could guarantee instead that anything like this would be removable.

So there it currently stands. I would love to find a solution that avoids the above problems while not adding too much complexity to the rest of Bril or to the SSA conversion algorithm itself. (The example is structured this way to closely mirror the idealized pseudocode; crapping it up too much just to deal with this weird edge case of undefined values would be a shame.)

Something in particular that I like about our current (flawed) approach is that it is possible to generate working SSA code with the "naive" SSA conversion algorithm; you don't need to do some sort of fancy "minimal SSA" conversion just to get correct code. It would be awesome to preserve that property in the face of undefined values.

One option I'm considering is redefining the `phi` instruction to explicitly support cases where the value is undefined. Then, it may be possible to avoid generating copies for undefined cases altogether. This may not work because of the need to *transitively* eliminate undefined copies, but maybe it will make the undefined values easier to clean up using trivial DCE.

Issue #120, opened 2021-01-12
[DF] defined analysis producing wrong output.


cgyurgyik:
### Example
```
# bril2json < p.bril | python3 ../../df.py defined

@main(awesome_integer: int) {
.entry:
  print awesome_integer;
}
```

### Actual:
```
entry:
  in:  ∅ 
  out: ∅
```

### Expected
`awesome_integer` should be defined in `.entry`.
```
entry:
  in:  awesome_integer
  out: awesome_integer
```

I think the issue here is two-fold.
1. There's no simple way to pass in arguments to the analyses. Here, we pass in an analysis with a defaulted (empty) data structure. However, for an analysis such as reaching definitions or defined analysis, we want to pass in the function arguments to the first block.
https://github.com/sampsyo/bril/blob/dbbd8e8f1d7000fefdcffaef2d79ce3339e28b40/examples/df.py#L82-L92

2. Even if we do pass in the correct function arguments,
https://github.com/sampsyo/bril/blob/dbbd8e8f1d7000fefdcffaef2d79ce3339e28b40/examples/df.py#L48-L49

```
inval = analysis.merge(out[n] for n in in_edges[node]) # in_edges[.entry] == []
```
So, even though `in_[.entry]` is initialized correctly, it is over-written afterward.

As usual, let me know if I'm off-target.

sampsyo:
Yeah, I wrote the DF analysis examples before function parameters existed. The right thing to do is to still use empty sets as the initial values but essentially hallucinate an entry block that defines the parameters.

PR #121, opened 2021-01-13
Add function parameters to data flow analysis.

cgyurgyik:
Addresses #120 .


sampsyo:
Seems possibly reasonable to literally generate "fake" code for the hallucinated block… but it could be even simpler, I imagine to instead do it as part of the algorithm? The tricky thing, I think, is that it might require a new "hook" into the DF framework, i.e., different analyses may want to treat argument differently. For example, an interval analysis will want to treat the parameters as unknown values. I'll stare at it sometime too!

cgyurgyik:
> Seems possibly reasonable to literally generate "fake" code for the hallucinated block… but it could be even simpler, I imagine to instead do it as part of the algorithm? The tricky thing, I think, is that it might require a new "hook" into the DF framework, i.e., different analyses may want to treat argument differently. For example, an interval analysis will want to treat the parameters as unknown values. I'll stare at it sometime too!

Yeah I think I was a bit narrow-minded when writing this (also, trying to keep the PR as small as possible), thinking only about a few DF algorithms such as reaching definitions and "defined" analysis.

PR #127, opened 2021-02-14
implement Display for rust types

rethab:
this adds a default `Display` implementation for the rust types formatting bril code in the text representation.

not sure if this is desired/helpful, but I find myself wanting to print something readable to the console again and again :)

### DIFF
--- a/bril-rs/.gitignore
+++ b/bril-rs/.gitignore
@@ -1,2 +1,3 @@
 **/target
 Cargo.lock
+.idea/**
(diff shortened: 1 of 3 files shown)


sampsyo (approved):
Seems like a great idea to me! Any objections, @Pat-Lafon?

Pat-Lafon:
This all looks good and is a nice addition.

Since it looks like the syntax of the `Display` is very close to the output of `bril2txt`, it may be easier/more thematic to test using turnt instead of rust-style tests. But this is something I will do in a separate pr.

sampsyo:
Excellent!! Thanks to both of you.

PR #128, opened 2021-02-15
bril2txt with Rust!

Pat-Lafon:
This builds upon #127 to create an example for `bril-rs` that trivially implements `bril2txt`. This allows the `Display` trait to be compared against the text version of a Bril program using turnt. Through this, I found two inconsistencies between the `Display` trait and `bril2txt` which have been fixed.

I have removed the original testing in favor of turnt + `make test`. This seemed a lot easier than managing two testing systems, the idomatic rust testing framework and the turnt framework.

Two more tests have been added to cover more features of `bril2txt`.

### DIFF
--- a/bril-rs/Cargo.toml
+++ b/bril-rs/Cargo.toml
@@ -10,12 +10,23 @@ edition = "2018"
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 
-[dev-dependencies]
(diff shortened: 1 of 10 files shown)


rethab (approved):
nice :tada:

sampsyo (approved):
Excellent! Nifty idea, @Pat-Lafon. :smiley:

PR #125, opened 2021-02-04
Brilirs stable

Pat-Lafon:
With Rust 1.51, `#![feature(or_patterns)]` should be stablized. This pr moves `brilirs` back on to the stable channel. This pr is dependent of #122.

See https://forge.rust-lang.org/ for the rust release schedule. The current date is March 25th.

PR #131, opened 2021-06-17
Bril-rs/Brilirs clippy warnings

Pat-Lafon:
Clippy is the builtin rust linter which is run with `cargo clippy`. This pr addresses Clippy warnings, most of which are about stylistic consistency and smoothing over a couple snippets using references.

Builds upon #125


sampsyo:
Nice!! Turning this on definitely seems like the thing to do. :+1: :sparkles:

PR #136, opened 2021-10-21
Bumping Rust edition

Pat-Lafon:
Today Rust released its 2021 edition alongside Rust 1.56 (https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html). This pr makes this edition change and requires no rust source code changes to `bril-rs` nor `brilirs`.

In checking this change(`cargo fix --edition`), it noticed that the old testing framework using `cargo test`(which was replaced with `turnt` and called via `make test`) was not maintained and no longer compiles. This code is obsolete so this pr also removes `brilirs/testdata/*` and `brilirs/tests/interp_test.rs`.

Thanks
Live from OOPSLA


sampsyo:
Hello, Chicago! I hope it's going great.

This all looks good; thank you!!

PR #137, opened 2022-01-17
Brilirs todo's and bril2json in Rust

Pat-Lafon:
This pr addressing some nagging todos that I've decided to cleanup:
- aef10d8 converts the `InterpError` enum to use the `thiserror` crate. This (to my understanding) is the standard rust crate for this type of error and in my opinion greatly improves it's readability.
- d35e03f cleans up some code.
- abf0768 and e4aa9ed updates the `clap` crate to 3.0. This deprecates the macro that was being used to create the cli for `brilirs` in favor of a derive macro on a `Cli` struct. In my opinion this is an improvement over the previous macro. This is also leveraged with the subcrate `clap_complete` to derive tab completions for a couple common shells like bash and fish.
- e7ee57b is my revived attempt at implementing `bril2json` in Rust. I don't love any of the parser generators in the Rust ecosystem but I've ended up using `lalrpop` which uses a not so great/cryptic `bril_grammr.lalrpop` grammar file. The original implementation is tested against `test/parse/` so this implementation does as well. Further, I've add the `--text` flag to `brilirs` which uses `bril2json` to let the interpreter take the text version of bril files directly. All of the `brilirs` tests have been updated to use `--text` so that both tools can be tested at the same time.
- f771749 in testing the previous commit, I found that one of the parsing test cases had a small error. The rust implementation of bril2json is more strict as it checks for valid operation names.
- 655c3d1 updates the documentation for `brilirs`, its tab completions, and `bril-rs` including installation instructions and the help page from `brilirs --help`.
- 361bb8e updates the `brilirs` benchmarks to also use `--text` and two bugs were fixed. The one wrinkle is that bril doesn't reserve any of it's operations as keywords so they can be used as identifiers. For most cases I've worked around this but I'm not sure if I can solve the reduce-reduce conflict with "const". I tried a solution, that worked, but it lead to the biggest blowup in rust build times that I've ever seen. I've decided that using "const" as an identifier is probably rare enough that it's fine to not handle this edge case. Maybe a different parsing crate would handle this better.

All tests should be runnable by calling `make test` in `brilirs/` and `bril-rs`. Benchmarks with `make benchmark`.
Note that running the tests requires building `turnt` from source for the 1.6.0 changes.


docs/tools/brilirs.md
```diff
@@ -11,16 +11,41 @@ Install
...
+        ARGS:
+            <ARGS>...    Arguments for the main function
+
+        OPTIONS:
+            -c, --check          Flag to only typecheck/validate the bril program
+            -f, --file <FILE>    The bril file to run. stdin is assumed if file is not provided
+            -h, --help           Print help information
+            -p, --profile        Flag to output the total number of dynamic instructions
+            -t, --text           Flag for when the bril program is in text form
+            -V, --version        Print version information
```
sampsyo, line 46: Instead of reproducing `brilirs --help` here, maybe it would be more useful (and easier to keep up to date) if we just tell people to run this command themselves? If there's anything we want to expand on about any of these CLI flags, we could just do that in prose here.

docs/tools/rust.md
```diff
@@ -17,12 +17,18 @@ features = ["ssa", "memory", "float", "speculate"]
...
+There are two helper functions: `load_program` will read a valid Bril program from stdin, and `output_program` will write your Bril program to stdout. Otherwise, this library can be treated like any other [serde][] JSON representation.
 
-Examples
+Tools
 ---
 
-There is currently a very trivial example that uses this interface to create a rust implementation of `bril2txt`. Run it with `cargo run --example bril2txt`.
+This library supports fully compatible Rust implementations of `bril2txt` and `bril2json`.
+
+For ease of use, these tools can be installed and added to your path by running the following in `bril-rs/`:
```
sampsyo, line 27:
```suggestion
For ease of use, these tools can be installed and added to your path by running the following in `bril-rs/`:

```

(To make the text below into a Markdown code block.)

sampsyo (commented):
Super cool; thanks for all of this stuff! These look like solid improvements; I'm especially happy with the error infrastructure improvements and the clap macro update.

About the new integrated parser: I know this is a natural choice given the interpreter's internal program representation, but I think it's somewhat limiting that the concrete syntax needs to bake in specific instruction and type names. This makes it a bit harder, for example, to use the parser alone (i.e., through the text-to-JSON conversion tool) when developing a new extension that the concrete-syntax grammar doesn't know about.

While it's out of scope for this PR, I guess I'd recommend a future course of action that uses two program representations: the current one (with `enum`s for the specific supported opcodes and such) and a new AST that more directly represents arbitrary Bril syntax (where opcodes are just plain strings). `bril2json` would use only the latter AST; `brilijs --text` would first load up this AST and then convert it to its more efficient internal representation. Just something to consider!

docs/tools/brilirs.md
Pat-Lafon: Resolved in 5d4d16c

Pat-Lafon:
I've create an AbstractProgram struct which I think is what you were getting at. I've switch both `bril2json` and `bril2txt` to this more relaxed bril representation. Notice how in `abstract_program.rs`, `AbstractInstruction` takes a string for its op field. `AbstractType` is similarly string based.

sampsyo:
Oh wow, super cool!! I like how this makes the `lalrpop` grammar so much simpler. Awesome work as always, @Pat-Lafon. 😊🚀

PR #138, opened 2022-01-23
Bril struct extension

mdmoeller:
Adds a Bril->LLVM compiler which supports the struct extension to a new examples/struct-ext.

Note that this change modifies bril-txt/briltxt.py to allow parsing of Bril programs that use structs.

Only existing documentation is my blog post:
https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/brilc

### DIFF
--- a/bril-txt/briltxt.py
+++ b/bril-txt/briltxt.py
@@ -17,7 +17,10 @@ __version__ = '0.0.1'
 # Text format parser.
 
 GRAMMAR = """
-start: func*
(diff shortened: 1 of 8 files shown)


sampsyo:
Nice; this all looks great! Great work, as ever.

I'll merge this now, because why not, but one thing we could consider doing in the future is adding proper documentation to the Bril docs about the struct extension. That would make it potentially easier for other people to build stuff on top of that extension.

PR #140, opened 2022-02-01
Add Armstrong benchmark and update docs

andrewb1999:
A bril program that determines if the input is an Armstrong number, a number that is the sum of its own digits each raised to the power of the number of digits.


sampsyo:
Excellent! Nifty idea. :smiley:

PR #141, opened 2022-02-02
primes-between benchmark

Yasgur99:
This PR introduces the `primes-between.bril` program and expected output `primes-between.out` when run with arguments `0 1000`.

This program accepts two arguments `a` and `b` and outputs to `stdout` the prime numbers in the interval `[a, b]`. It is assumed that `a, b >= 0`. and `a <= b`


sampsyo (changes requested):
Cool; looks good so far! I have some comments on the summary in the docs, and I think you may still need to commit the instruction-count file.

docs/tools/bench.md
```diff
@@ -33,6 +33,7 @@ The current benchmarks are:
 * `sqrt`: Implements the [Newton–Raphson Method][newton] of approximating the square root of a number to arbitrary precision
 * `sum-sq-diff`: Output the difference between the sum of the squares of the first *n* natural numbers and the square of their sum.
 * `binary-fmt`: Print the binary format for the given positive integer.
+* `primes-between`: Print the primes int the interval `[a, b]`.
```
sampsyo, line 36:
```suggestion
* `primes-between`: Print the primes in the interval `[a, b]`.
```

Looks overall good, but I think still missing the profiling output file? (Hence the failing CI.)

sampsyo (changes requested):

docs/tools/bench.md
Yasgur99: @sampsyo I wasn't sure how to run turnt to get the number of instructions for command that takes input. I posted in zulip under #bril about it

Yasgur99:
@sampsyo all changes requested should be good to go now.

sampsyo:
Looks like the tests are still failing. Can you maybe try running them locally to see what's wrong?

Yasgur99:
> Looks like the tests are still failing. Can you maybe try running them locally to see what's wrong?

@sampsyo 
If I run `turn -v primes-between.bril` the test fails but if I run `turnt -v -a "1 1000" primes-between.bril` then it passes. I'm assuming that the turnt config does not run my tests with these arguments. I'm not sure what kind of changes to make since other tests would not want this input to run the test. One thing I could do is hardcode the test to take no arguments and modify the bril code to create temps containing these values and use these instead. Do you know of another way to keep it accepting arguments?

sampsyo:
Aha! The way to do this is with an `# ARGS: ...` comment inside the file. Like this, in the recently-added `armstrong` benchmark:
https://github.com/sampsyo/bril/blob/825a690283bff1e4781a99e1d164c1456cb61a8f/benchmarks/armstrong.bril#L1

Here are [the relevant Turnt docs](https://github.com/cucapra/turnt#details).

PR #142, opened 2022-02-02
Added benchmark: print factors of a number

anshumanmohan:


anshumanmohan:
@sampsyo, I didn't know a good place to give attributions but I'd like to note that I referred to the Ackermann and GCD benchmarks to get a sense of bril formatting and argument passing. Sorry if that was less than kosher; please let me know and I'll go about it differently next time.

sampsyo:
Awesome! No, that's perfectly legit—looking at existing code is a great way to learn how any language works. :smiley: I think it's important to avoid looking at much other code when you're working through an actual compiler algorithm (as you'll do soon) to force yourself to work through it, but figuring out how to write Bril programs really is best done by example.

benchmarks/factors.bril
```diff
@@ -0,0 +1,35 @@
...
+  print fac;
+  num: int = div num fac;
+  jmp .loopcheck;
+
+.ifno:
+  fac: int = add fac one; 
+  jmp .loopcheck;
+
+.loopend:
+  ret one;
```
Pat-Lafon, line 34: Brilirs rejects this which I believe is correct based on #91. Functions without a return type should not return a value.

anshumanmohan: Thanks, I think I've resolved this [here](https://github.com/sampsyo/bril/pull/150).

PR #144, opened 2022-02-02
Add Benchmark: Bubble Sort for 5 elements

tonyjie:


sampsyo:
Excellent! Pretty nifty benchmark; nice work.

PR #146, opened 2022-02-02
Add Benchmark: Adj2CSR

chhzh123:
This PR introduces a new benchmark called `adj2csr`, which converts a random graph in adjacency matrix format to [Compressed Sparse Row (CSR)](https://en.wikipedia.org/wiki/Sparse_matrix) format.

The random graph generator reuses the code in `benchmark/mat-mul.bril`. Basically, users can input number of nodes and a fixed seed to the program, and the program will output the adjacency matrix, the CSR offset array, and the CSR edges array.


sampsyo:
Very cool! Nice use of the memory extension.

PR #147, opened 2022-02-02
Added Adler-32 benchmark.

5hubh4m:


sampsyo:
Super interesting! Nice work!

5hubh4m:
Writing this benchmark made me think we should have bitwise ops in the spec haha!

sampsyo:
Indeed; the `@bitwise_or` function is a pretty good demonstration of that. 😂

benchmarks/adler32.bril
```diff
@@ -0,0 +1,84 @@
...
+}
+
+# Fills the array with values 0, 1. 2, ...
+@fill_array(arr: ptr<int>, size:int) {
+  one: int = const 1;
+  curr: int = const 0;
+  loc: ptr<int> = id arr;
+.loop:
+  store loc curr;
+  loc: prt<int> = ptradd loc one;
```
Pat-Lafon, line 25: From brilirs: `prt` -> `ptr`

5hubh4m: Interesting that `brili` didn't catch this.

Pat-Lafon: My understanding is that `brili` does it's type checking based on the type of the value being evaluated and ignores the destination type.

sampsyo: Very true; brili is purely dynamic. There are static type checkers elsewhere in the repository.

Issue #148, opened 2022-02-02
[Interpreter] Store instruction doesn't work with float pointer


zzzDavid:
## Description

`store` instruction raises error when writing to a float pointer.

### Example

```
@main() {
  one: int = const 1;
  seed: float = const 109658.0;
  rng: ptr<float> = alloc one;
  store rng seed;
}
```

### Error message
```
../bril/bril-ts/build/brili.js:789
process.on('unhandledRejection', e => { throw e; });
                                        ^
store argument 1 must be a Pointer
```

sampsyo:
Thanks!! The problem in the reference interpreter is here:
https://github.com/sampsyo/bril/blob/9fa84eb485b16129a35d5e45bab4ca02dad57d6f/bril-ts/brili.ts#L632-L645

That case for `store` doesn't need to be a switch; it should just use `getArgument` directly:
https://github.com/sampsyo/bril/blob/9fa84eb485b16129a35d5e45bab4ca02dad57d6f/bril-ts/brili.ts#L239

…and support any `target.type`.

Pat-Lafon:
@sampsyo I believe this issue can be closed.

sampsyo:
Indeed; thanks!

PR #149, opened 2022-02-02
Interpreter store instruction fix: issue 148

zzzDavid:
### Description

This pull request is to solve issue: https://github.com/sampsyo/bril/issues/148.

### Changes
Changed the get argument switch case in `store` instruction to `getArgument` function to support any-type store.


bril-ts/brili.ts
```diff
@@ -629,20 +629,7 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
...
-      case "bool": {
-        state.heap.write(target.loc, getBool(instr, state.env, 1));
-        break;
-      }
-      default: {
-        state.heap.write(target.loc, getPtr(instr, state.env, 1));
-        break;
-      }
-    }
+    state.heap.write(target.loc, getArgument(instr, state.env, 1));
```
sampsyo, line 632: Consider passing `target.type` as the fourth argument to `getArgument` to enable type checking.

sampsyo (changes requested):
Thanks for getting this started so quickly!!

It looks like the test `wrong_write.bril` is failing because we no longer throw an error when trying to store a `bool` value into a `ptr<int>` pointer. Adding some type checking back to the `getArgument` call would help.

It would also be extra-awesome to add your short test program from #148 as a new test so we can be sure that `ptr<float>` is now working in the interpreter.

bril-ts/brili.ts
zzzDavid: Thanks! I just added that, it reminds me always check types before writing ;)

sampsyo:
Cool—what do you think about the above suggestion to include your test case?

zzzDavid:
> Cool—what do you think about the above suggestion to include your test case?

Thanks, just added that

sampsyo (approved):
Looks perfect; thanks!

PR #152, opened 2022-02-03
Aaron/benchmark

atucker:
This PR adds an implementation of hyperoperations to the benchmarks, expressed using Knuth's up arrow notation. Hyperoperations generalize the way that multiplication is repeated addition. This adds exponentiation, but also repeated exponentiation, and repeated repeated exponentiation, etc.

https://en.wikipedia.org/wiki/Knuth%27s_up-arrow_notation


sampsyo:
Very nifty idea! Looks great, and potentially very expensive. :smiley:

benchmarks/up-arrow.bril
```diff
@@ -0,0 +1,36 @@
...
+}
+
+@up_arrow(num: int, arrows: int, repeats: int): int {
+  one: int = const 1;
+
+  ans: int = id num;  
+  i: int = const 1;
+
+.loopstart:
+  keepgoing: int = lt i repeats;
```
Pat-Lafon, line 15: `int` -> `bool`?

atucker: Oh, good catch! Yeah brili still lets this work for some reason...

Pat-Lafon: Yeah, different implementations. I'm just trying to make sure `brilirs` runs on all of the new benchmarks that are being added and its more strict about these things.

PR #153, opened 2022-02-03
This benchmark calculates the inverse of a 3x3 matrix

charles-rs:
It is pretty memory efficient, as it only allocates the matrix and a destination for the inverse. some of the logic for indexing into the matrices would benefit greatly from some CSE...


charles-rs:
forgot to edit docs; ignore

PR #154, opened 2022-02-03
Add benchmark for inverting matrix

charles-rs:
inverts a 3x3 matrix. The indexing could benefit from some CSE


sampsyo:
Very cool, and potentially even realistic. And leaving some work to do for some future optimizations seems like a good idea too. 😃

Issue #155, opened 2022-02-03
Add type checker to GitHub Actions


sampsyo:
Most of the tests we run in CI rely on the reference interpreter, which is (by design) extremely permissive. Plenty of type errors are forgiven, for example. As @Pat-Lafon has valiantly pointed out lately, however (e.g., https://github.com/sampsyo/bril/pull/154#discussion_r798688341 and https://github.com/sampsyo/bril/pull/152#discussion_r798671074), this makes it easy for hand-written code with type errors to sneak into the benchmarks and other tests.

It would be awesome to set up automated checking in CI for all the Bril programs we have in tree. We could do this with [brilirs](https://capra.cs.cornell.edu/bril/tools/brilirs.html)'s own `--check` mode or the [type inference tool](https://capra.cs.cornell.edu/bril/tools/infer.html).

Bonus points for pointing out the line where the error occurs and adding a [check annotation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks) on the offending line in a given PR… but that can come in a subsequent phase!

PR #157, opened 2022-02-03
rectangles benchmark

ayakayorihiro:
New benchmark for bril for CS6120 Lesson 2 tasks; this benchmark takes the side values of two rectangles, and computes the difference in area between the two rectangles.


sampsyo:
Looks perfect; thanks!!

benchmarks/rectangles-area-difference.bril
```diff
@@ -0,0 +1,19 @@
+# ARGS: 5 10 6 13
+@main (x1 : int, y1: int, x2: int, y2 : int) {
+      a1: int = call @area x1 y1;
+      a2: int = call @area x2 y2;
+      res: int = sub a1 a2;
+      a1_bigger: boolean = gt a1 a2;
```
Pat-Lafon, line 6: Did you mean: `boolean` -> `bool`?

ayakayorihiro: Ahh sorry, it seems like I missed the syntax a bit. But weirdly enough `boolean` also works? (I tested both ways locally) I'll make a PR with the fix, thanks for pointing this out!

Pat-Lafon: Yeah it's just #155. It's technically fine for `brili` but I've been trying to keep `brilirs` working on all of the benchmarks and it more strictly follows the specification.

PR #158, opened 2022-02-03
[Benchmark] Fix type issue for cholesky benchmark

zzzDavid:
This PR is a minor fix for some type issues in the Cholesky benchmark. Thanks to @Pat-Lafon for pointing them out.


sampsyo (approved):
Looks like it's headed in the right direction to me!

PR #160, opened 2022-02-03
Bril Benchmark-Riemann Sums

JonathanDLTran:
Create benchmark for Bril with left, midpoint and right Riemann Sum calculations, using the function f(x) = x^2.


docs/tools/bench.md
```diff
@@ -64,3 +65,4 @@ Credit for several of these benchmarks goes to Alexa VanHattum and Gregory Yaune
 [armstrong]: https://en.wikipedia.org/wiki/Narcissistic_number
 [adler32]: https://en.wikipedia.org/wiki/Adler-32
 [uparrow]: https://en.wikipedia.org/wiki/Knuth%27s_up-arrow_notation
+[riemann]: https://en.wikipedia.org/wiki/Riemann_sum#:~:text=In%20mathematics%2C%20a%20Riemann%20sum,integral%20by%20a%20finite%20sum.&text=One%20very%20common%20application%20is,of%20curves%20and%20other%20approximations.
```
sampsyo, line 68:
```suggestion
[riemann]: https://en.wikipedia.org/wiki/Riemann_sum
```
Maybe a link to the article itself suffices?

sampsyo (commented):
Awesome; looks great! The link to Wikipedia seemed a tad odd, though; is the shorter one OK?

docs/tools/bench.md
JonathanDLTran: Looks like I made a mistake on the Wikipedia article, and the shorter one is correct.

sampsyo:
Great; thanks!!

benchmarks/riemann.bril
```diff
@@ -0,0 +1,89 @@
...
+    br b .while.end .while.body;
+    .while.body:
+    offset : float = fmul delta i; 
+    x : float = fadd a offset; 
+    f_x : float = call @square_function x;
+    sum = fadd sum f_x;
+    i = fsub i one;
+    jmp .while.header;
+    .while.end:
+    sum = fmul sum delta;
```
Pat-Lafon, line 39: `: float`? This seems to hold true for each of the riemann functions.

sampsyo: If you want, @JonathanDLTran, the [type inference](https://capra.cs.cornell.edu/bril/tools/infer.html) tool can automatically fill these in!

JonathanDLTran: I will definitely be using the type inference tool!  I forgot to use it when I was writing this, but it would save a lot of time.

PR #163, opened 2022-02-04
[Benchmark] Add binary search

yy665:
A not-very-interesting binary search implementation.


sampsyo (changes requested):
Looks like a good start! The tests seem to be failing, probably because of a missing newline on the `.out` and `.prof` output files. Can you try using `turnt --save` to update those and see if that fixes it?

sampsyo:
Hi, @yy665—just a reminder that this needs a second look!

yy665:
> Hi, @yy665—just a reminder that this needs a second look!

Sorry, I just saw the notification. I have pushed the fix.

sampsyo:
Hmm; looks like there are now *two* newlines at the ends of both files? Using `turnt --save` could bypass this trouble, as opposed to manually editing the files…

sampsyo (approved):
Looks like everything's green now!

PR #165, opened 2022-02-04
fixed boolean to bool

ayakayorihiro:
Somehow it seems like `boolean` also worked, but the proper syntax is `bool`. Sorry about that!


sampsyo:
Thanks for the quick fix! Yep, `brili` does not do type checking.

ayakayorihiro:
Ahh gotcha, sorry for the bug!!!

PR #167, opened 2022-02-04
Adding positions support to bril-rs and brilirs

Pat-Lafon:
- c3965bf Adds feature-gated support for positions to bril-rs. Adds `make features` to do a naive check that all features are properly gated. Adds `make install` for the rust versions of `bril2txt` and `bril2json`. Small updates to documentation.
- 34bc048 Adds positions to the rust version of `bril2json` with the `-p` flag. Updates test config to take arguments.
- 16011b9 Adds positions to `brilirs` for more information in error reports when available.
- 1c224c2 and bd45453 Add support to `brilirs` for returning errors when the type signature is missing.

Programs that are ill-formed such that they break parsing will not return nice error messages. `brilirs` should return a position with an error when possible but there are not currently tests for this.

Examples:
```sh
$ bril2json -p < ../benchmarks/mat-inv.bril | cargo run

error: Line 54, Column 3: Missing type signature
```

```sh
$ bril2json -p < ../benchmarks/up-arrow.bril | cargo run -- 2 3 3

error: Line 15, Column 3: Expected type `Bool` for assignment, found `Int`
```

### DIFF
--- a/bril-rs/Cargo.toml
+++ b/bril-rs/Cargo.toml
@@ -16,6 +16,7 @@ float = []
 memory = []
 ssa = []
 speculate = []
+position = []
 
 [[example]]
 name = "bril2txt"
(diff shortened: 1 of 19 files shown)


sampsyo (approved):
Wow; looks super cool!! So fast! And nice work sorting out the position rigamarole in lalrpop. :clap: :clap: :clap:

PR #171, opened 2022-02-07
Fixed an lvn test by removing the invalid ne instruction.

5hubh4m:


sampsyo:
Awesome; thanks for fixing this!!

PR #175, opened 2022-02-10
bril-rs Housekeeping and Github actions for Rust

Pat-Lafon:
I was separately thinking about a clippy lint for the simpler 2021 edition format arguments like in 7432319 when I realized that I had a small `bril2json` bug in 2ca7f95. With the recent work in support of #155, I was inspired by this bug to suggest a similar set of github actions for running basic checks(`check`, `fmt`, and `clippy`) on the rust code bases in 62cb25c. I've never really used github actions before, but `actions-rs` seems to be leading project for rust actions and with some modification to the code base in f2bc178 to support this, I was pleasantly surprised that it found 8e51666 that I had missed when fiddling with the clippy lints in df247bf. 

One could run more checks through the workflow like running `brilirs --check` against the benchmarks but with the development of `brilck` this is probably becoming less necessary and also more likely to become outdated compared to the main bril toolchain.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -1,0 +1,39 @@
+on:
+  push:
+    branches:
+     - main
+  pull_request:
+    branches:
+     - main
+
+name: Workflow checks for rust code
+
+jobs:
+  rust-ci:
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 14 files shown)


sampsyo:
Awesome! Seems to work, AFAICT!

PR #176, opened 2022-02-11
Try adding brilck to CI

### LINKED ISSUE #155: Add type checker to GitHub Actions

sampsyo:
This would, at last, fix #155! Probably needs some work though.


sampsyo:
OK, I've tried this with an intentional error and it seems to correctly annotate the right line of the right file with a type error. :tada:

PR #178, opened 2022-02-13
Add catalan number benchmark

susan-garry:
Recursively compute the nth [Catalan number](https://en.wikipedia.org/wiki/Catalan_number).


benchmarks/catalan.bril
```diff
@@ -0,0 +1,33 @@
...
+  idx: int = id zero;
+  n: int = sub n one;
+.while:
+  guard1: bool = le idx n;
+  br guard1 .while.body .while.end;
+.while.body:
+  n2: int = sub n idx;
+  v1: int = call @catalan idx;
+  v2: int = call @catalan n2;
+  elti = mul v1 v2;
```
sampsyo, line 27: Looks like this is missing a type annotation.

sampsyo (changes requested):
Looks good overall! But brilck is complaining that there is a missing type annotation in the benchmark code.

benchmarks/catalan.bril
susan-garry: Should be fixed now!

sampsyo (approved):
Looks like it's all set!

PR #179, opened 2022-02-25
Benchmark causing error

Andrey-Yao:
One of the benchmarks writes "boolean" instead of "bool" which is making my json parser sad


sampsyo:
Hello! Thanks for the PR, but I think this was already fix on the `main` branch (which is why there is a "null diff" under the "Files changed" tab here in the PR).

PR #180, opened 2022-03-11
Rust Library and Brilirs documentation

Pat-Lafon:
This PR addresses the lack of documentation for the Rust code bases in Bril and some clippy lints.

- 2287ded Now enforcing the `clippy::cargo_common_metadata` lint. This ensures that all of the important metadata fields in `Cargo.toml` are filled out. The important thing to note is the `license-file` field which I've pointed towards the `LICENSE` file in the root of the repository.
- 3c0ec80 Just mentions that `bril_rs` supports source positions in the rust library page. There is also a minor change where "position" was not added as a feature field in the `brilirs` dependency on `bril_rs`. This did not break the build because features are additive in Rust so it was included by the `bril2json` dependency instead during dependency resolution.
- c0a7be4 In recent minor releases of clap, some methods were deprecated and are addressed here.
- 740b440 The driving commit of this pr. While there is documentation of these code bases on the Bril website, its probably not sufficient. This pr leverages `rustdoc`, the builtin documentation support for Rust, to document the higher level design of these code bases. In each directory, one can use `cargo doc --open` to get very basic documentation on the public functions and data structures provided by `bril_rs`, `brilirs`, and `bril2json`. `cargo doc` is also added to the github workflow alongside the other rust actions.

The documentation itself is not detailed however, I think it accomplishes the goal of elevating the useful functions/interfaces of these libraries from the implementation. The expectation is that someone looking to use one of these libraries could use `cargo doc --open` to get exposed to the public functionality of these libraries without getting overwhelmed by the source code. Feedback is appreciated.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -49,3 +49,8 @@ jobs:
         with:
           command: clippy
           args: --manifest-path ${{ matrix.path }} -- -D warnings
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --manifest-path ${{ matrix.path }}
--- a/bril-rs/Cargo.toml
+++ b/bril-rs/Cargo.toml
@@ -3,6 +3,13 @@ name = "bril-rs"
 version = "0.1.0"
 authors = ["Patrick LaFontaine <32135464+Pat-Lafon@users.noreply.github.com>"]
 edition = "2021"
+description = "A rust representation of the Bril language"
+readme = "README.md"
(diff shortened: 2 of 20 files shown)


sampsyo:
Awesome!! I read over all the docs and I think this is a great start. I like the tactic of relying on links to the Bril language reference to explain what's going on with the various parts of the program representations.

I'm going to merge this now since it's already quite useful. Folks using the Rust library (@andrewb1999?) can weigh in about things they wish were documented in more detail!

PR #184, opened 2022-04-02
convert missing label panic to InterpError

Pat-Lafon:
Currently `brilirs` panics if a Bril program tries to jump to label that doesn't exist. This pr converts this panic to an InterpError::MissingLabel.


sampsyo:
Awesome; I am a big fan of producing errors!! 🎉

PR #185, opened 2022-04-02
seperate workflow into docs.yaml

Pat-Lafon:
Currently on every push, the Github workflow tries to generate/deploy the docs. For any fork of bril, this job will always fail. In the same vein as #175, this pr moves the documentation generation into its own workflow that only runs when a file is modified in `docs/**`. This was moved into a separate workflow as it seems nontrivial to filter by `path:` for specific jobs.


sampsyo:
Seems mostly good, but can we remove the `branches: [main]` restrictions? It would be great to build the docs on all branches including pull requests, so it's possible to detect errors in the docs before merging a PR.

There is already a restriction that only actually does the rsync deployment step if we're on the `main` branch:

```
if: ${{github.event_name=='push' && github.ref=='refs/heads/main'}}
```

sampsyo:
Looks great; thank you!!!

PR #186, opened 2022-04-02
Allow profiling output location to be provided

Pat-Lafon:
This is a minor quality of life change to `brilirs` to allow the user to specify the output location of profiling. 

This will allow a user to execute their program and send the results to `my_program.out` and `my_program.prof` respectively when using `brilirs` as a library.


sampsyo:
Nice; all looks good to me!!

PR #187, opened 2022-04-02
remove unsafe code from brilirs

### LINKED ISSUE #183: Rust code doesn't conform to commonly accepted `unsafe` semantics
Hi,

So I'm not very familiar with the project, but I've done a fair bit of Rust and one of my friends is currently taking 6120, so I looked a little bit through the code. It seems some of the Brilrs code isn't using the commonly accepted Rust definitions of unsafe, which means that one could accidentally run into undefined behavior. Now these issues aren't actually being realized at the moment, but it might be good to conform to Rust standards to ensure correctness in the future.

The commonly accepted Rust standard is to say that

> A function should be marked as `unsafe` if it's possible to cause undefined behavior by calling it from safe code.

If one were to do (in entirely safe code)

```rs
let bbprog: BBProgram = prog.try_into()?;
interp::execute_main(&bbprog, out, &input_args, profiling)?;
```

one could actually cause undefined behavior if the program passed into `execute_main` is not a valid program. This is because (if I'm reading correctly) `execute_main` assumes the program has been type-checked
https://github.com/sampsyo/bril/blob/30c836cb208720a0060dfa30480658f7af31a20c/brilirs/src/interp.rs#L184-L218
which is not enforced by the Rust compiler. Thus one can cause undefined behavior in entirely safe code, which is against Rust conventions. As I said, this is of course not a *real* problem and it doesn't currently ever materialize itself, but it might be nice to write better Rust.

There's a couple possible solutions to this, if it is worth solving:

1. Have `type_check` return some sort of `TypedProgram`, and `execute_main` can only take in a `TypedProgram`, guaranteeing that the program has been type-checked. This solution makes the most sense to me.
2. Mark `execute_main` as unsafe. This is probably the least work but is very sad.
3. An unreasonable solution: rework nodes/type-checking so that type-checking returns a new tree that doesn't even have the nodes that should be unreachable. This is very Haskell-y and a huge pain to do.

Pat-Lafon:
I would like to resolve #183 by removing the use of `unreachable_unchecked` from `brilirs`. I am suggesting this for two reasons. 

First, while one could use `TypedBBProgram` to enforce type checking before interpretation, it's actually not obvious to me that the current type checking and validation is sufficient to avoid undefined behavior. Well-formedness as defined in the Bril specification is a dynamic property where as a `TypedBBProgram` would be only statically checked. Given this mismatch, I'm not sure if there are Bril programs, well formed or not, that could cause undefined behavior in `brilirs`. In practice, this hasn't been an issue since all of the Bril programs I've seen are limited enough to be statically checked.

Second, I've revisited the performance benefits of using unsafe code in `brilirs`. It appears that using the current benchmarks, on my noisy setup, that the performance gain of `unreachable_unchecked` is just a few percent on the slowest benchmarks(which puts it within error). For some further ad-hoc validation, I wrote a Bril program(not included) that just looped and added 2 numbers for about 300ms. To my surprise, it actually had less of an impact(less absolute time saved) than on `ackermann.bril` which runs for about 15ms. I speculate that this is because branch prediction is so consistent that it can just guess the pattern match right almost all of the time.

Given this, I think it's probably better just to remove the unsafe code from `brilirs`.

### DIFF
--- a/brilirs/src/interp.rs
+++ b/brilirs/src/interp.rs
@@ -1,5 +1,4 @@
 use std::fmt;
-use std::hint::unreachable_unchecked;
 
 use crate::basic_block::{BBFunction, BBProgram, BasicBlock};
 use crate::error::{InterpError, PositionalInterpError};
@@ -153,8 +152,7 @@ impl fmt::Display for Value {
       Value::Bool(b) => write!(f, "{b}"),
       Value::Float(v) => write!(f, "{v}"),
       Value::Pointer(p) => write!(f, "{p:?}"),
-      // This is safe because Uninitialized is only used in relation to memory and immediately errors if this value is returned. Otherwise this value can not appear in the code
-      Value::Uninitialized => unsafe { unreachable_unchecked() },
+      Value::Uninitialized => unreachable!(),
     }
   }
 }
@@ -187,8 +185,7 @@ impl From<&Value> for i64 {
     if let Value::Int(i) = value {
       *i
     } else {
-      // This is safe because we type check the program beforehand
-      unsafe { unreachable_unchecked() }
+      unreachable!()
     }
   }
 }
@@ -199,8 +196,7 @@ impl From<&Value> for bool {
     if let Value::Bool(b) = value {
       *b
     } else {
-      // This is safe because we type check the program beforehand
-      unsafe { unreachable_unchecked() }
+      unreachable!()
     }
   }
 }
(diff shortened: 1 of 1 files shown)


sampsyo:
Fantastic; this seems like a very clean win. Especially for the sake of debuggability, this seems like a nice place to be even if it _did_ come at a measurable-but-small performance penalty, which it seems not to.

PR #188, opened 2022-04-02
Convert environment into a stack

Pat-Lafon:
Each time a function is called in `brilirs`, a new environment represented as a vector is allocated. This is costly for functions that are heavily recursive, small, and make multiple function calls in their body. To show this, I've created a benchmark called `function_call.bril` which is an extreme version of this.

To optimize function calls, I've modified the `Environment` struct to act like a stack where there are stack pointers into a `Vec<Value>` representing the frames of the stack. In this way, there is only one `Environment` struct which grows in size on each recursive function call by calling `Environment::push_frame`. Likewise, when a function call returns, the frame is popped with `Environment::pop_frame` which allows for that space to be reused in the next function call.

Some rough numbers for each of the interpreters on `function_call.bril`.

| brili  | Old brilirs | Proposed brilirs |
| ------------- | ------------- | ---------|
| 7.2s ± .2 s  | 347 ms ± 4ms | 321 ms ± 3 ms |

In terms of the other benchmarks, `ackermann.bril` as the current benchmark with the most recursion sees a reduction from 15.2ms -> 13.2ms for brilirs.

### DIFF
--- a/benchmarks/function_call.bril
+++ b/benchmarks/function_call.bril
@@ -1,0 +1,22 @@
+# ARGS: 25
+@main(m: int) {
+  zero: int = const 0;
+  one: int = const 1;
+  two: int = const 2;
+  cond_m: bool = eq m zero;
+  br cond_m .end .m_nonzero;
+.m_nonzero:
+  m:int = sub m one;
+  call @main m;
+  cond_m: bool = eq m zero;
+  br cond_m .end .m_nonzero2;
+.m_nonzero2:
+  m:int = sub m one;
+  call @main m;
+  cond_m: bool = eq m zero;
+  br cond_m .end .m_nonzero3;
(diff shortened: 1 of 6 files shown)


sampsyo:
Neato!! I don't know if this is a coincidence or you happened to see the discussion, but in the GC task we just did for 6120, the desire for a "unified" environment just came up in https://github.com/sampsyo/cs6120/discussions/297#discussioncomment-2487294. I don't know how much easier this new environment would have made @andrewb1999's life, but maybe a _little_ easier?

Anyway, this seems like both a clarification to the data structures involved on top of the performance improvement. Seems great!!

andrewb1999:
This actually would have made my implementation much easier! This is not too far off from how I ended up implementing it actually, just a slightly more efficient data structure.

PR #191, opened 2022-04-20
Using source code in Brilirs error messages

Pat-Lafon:
I've felt inspired to bring together the work of #137 (which allows brilirs to handle Bril text files as input) and #167 (which implements source positions) in implementing error messages from brilirs which contain the source code of where error occurred(Inspired by Rust and other languages with nice error messages).

- fd969de adds an optional flag to the cli `--source/-s` which allows for an arbitrary source file to be provided for error messages. The only assumptions made on this source file is that each line is separated by a `\n` character and that the `col` position indicates the number of space characters into the line(meaning, I haven't considered tabs or other characters that could be only one character in terms of the `col` position of the source code but are multiple space characters in size). If this flag is not provided, but `--text/-t` is, then the input Bril text file will be assumed to be the source file. Source code will only be used in errors if there is an associated position for the error which might not be true if the provided Bril JSON file does not contain source positions or if brilirs does not assign a position for the error(for example, the `NoMainFunction` error).
- c20d39b fixes a crash when the function body is empty.

An example error message:
```
❯ brilirs --text --file test/interp-error/call-wrong-arity.bril 
error: Line 5, Column 5: Expected `2` instruction arguments, found `3`
    call @addboth x y z;
    ^
```

The errors/error messages in brilirs are not that well tested. I only found out about the edge case that caused c20d39b from running some for the `interp-fail` tests by hand. brilirs is currently not tested against the `interp-fail` and `check` tests but is tested against `fail` because the setup of both seem to be comparing against `*.err` files(note that brilirs error messages are not the same as brili or brilck error messages) whereas `fail` compares the output against `*.out` files(I'm not really sure why the test suite does this, all of the `*.out` files are empty). It seems like the minimum that should be done is to just check the error code brilirs gives to check that it provides errors/doesn't crash. I'm not sure if turnt will just let me diff error codes and not files or if I should create empty `*.out` files for `interp-fail` and `check`(or if there is a third, better option).

Brilirs error messages are best effort and haven't been check that thoroughly.

### DIFF
--- a/bril-rs/src/conversion.rs
+++ b/bril-rs/src/conversion.rs
@@ -43,11 +43,11 @@ pub enum ConversionError {
 impl ConversionError {
     #[doc(hidden)]
     #[must_use]
-    pub fn add_pos(self, pos_var: Option<Position>) -> PositionalConversionError {
+    pub const fn add_pos(self, pos_var: Option<Position>) -> PositionalConversionError {
         match self {
             //Self::PositionalConversionErrorConversion(e) => e,
             _ => PositionalConversionError {
-                e: Box::new(self),
+                e: self,
                 pos: pos_var,
             },
         }
@@ -57,18 +57,17 @@ impl ConversionError {
 /// Wraps [`ConversionError`] to optionally provide source code positions if they are available.
 #[derive(Error, Debug)]
 pub struct PositionalConversionError {
-    e: Box<ConversionError>,
-    pos: Option<Position>,
+    #[doc(hidden)]
+    pub e: ConversionError,
+    #[doc(hidden)]
+    pub pos: Option<Position>,
 }
 
 impl PositionalConversionError {
     #[doc(hidden)]
     #[must_use]
-    pub fn new(e: ConversionError) -> Self {
-        Self {
-            e: Box::new(e),
-            pos: None,
-        }
+    pub const fn new(e: ConversionError) -> Self {
+        Self { e, pos: None }
     }
 }
 
--- a/brilirs/src/basic_block.rs
+++ b/brilirs/src/basic_block.rs
@@ -279,6 +279,9 @@ impl BBFunction {
   }
 
   fn build_cfg(&mut self, label_map: FxHashMap<String, usize>) -> Result<(), InterpError> {
+    if self.blocks.is_empty() {
+      return Ok(());
+    }
     let last_idx = self.blocks.len() - 1;
     for (i, block) in self.blocks.iter_mut().enumerate() {
       // If we're before the last block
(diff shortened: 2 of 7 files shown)


sampsyo:
Very nice; this looks great!! I'm not sure if you're interested in taking another dependency, but there are a couple of widely-used crates out there that do the whole "point at a position in the source file" error reporting thing that rustc also does. Namely, [codespan](https://github.com/brendanzab/codespan) is one that just has to do with printing, and [miette](https://docs.rs/miette/latest/miette/index.html) is a more holistic solution with its own result traits and such.

About that directory of `fail` tests: those predate any static checking, so they are just "negative examples" meant to show that the interpreter fails at all in some ad hoc way. And yeah, all it does is check the exit code; it doesn't attempt to check anything about the error message. (That's what's up with the `return_code = 2` in the `turnt.toml`.) `turnt --diff` doesn't (currently!) say anything about error codes, but plain ol' `turnt` does that just fine (it tells you the expected and actual error code whenever they differ).

Pat-Lafon:
> Very nice; this looks great!! I'm not sure if you're interested in taking another dependency, but there are a couple of widely-used crates out there that do the whole "point at a position in the source file" error reporting thing that rustc also does. Namely, [codespan](https://github.com/brendanzab/codespan) is one that just has to do with printing, and [miette](https://docs.rs/miette/latest/miette/index.html) is a more holistic solution with its own result traits and such.

I didn't consider this approach because the source position extension doesn't use spans. Maybe this would be worth investigating further since these crates do seem to give nice messages. While it could be done without, is there any reason that would block extending #161 to contain a span instead of a single code point?

> About that directory of fail tests: those predate any static checking, so they are just "negative examples" meant to show that the interpreter fails at all in some ad hoc way. And yeah, all it does is check the exit code; it doesn't attempt to check anything about the error message. (That's what's up with the return_code = 2 in the turnt.toml.) turnt --diff doesn't (currently!) say anything about error codes, but plain ol' turnt does that just fine (it tells you the expected and actual error code whenever they differ).

To clarify, I understand that I can check the `return_code`. However, turnt doesn't just check the `return_code`, it also checks the output against some file. If you provide `output.err = "2"`, it checks against `*.err` files like in `test/check/`. If you leave this empty as in `test/fail/`, it checks the output against `*.out`. What I'm looking for is whether turnt supports some way of only doing the first part, the checking of the  `return_code` and skipping the second part, checking the output against some file(because my output of an error message won't match and checking against stdout would require adding a bunch of empty `*.out` files to `check`/`interp-error`).

sampsyo:
A span thing could be cool! I didn't do it at first just because it was easier to start this way and I wasn't sure anyone actually cared about the span for errors in an instruction-based language (where the position is pretty unambiguous anyway). :smiley:

I'm not 100% sure, but maybe `turnt` could be configured to check no outputs at all by just setting the entire `output` option in the config file to be empty? Like, `output = {}` might be the relevant TOML syntax.

PR #192, opened 2022-04-26
Bitwise operators for XOR, AND, and OR

thomasyang18:
The algorithm is using multiplications and divisions to simulate bit shifts, although while normally one might implement it via something like x&(1<<i) for every bit to check whether that bit is on or not, I keep left shifting A and B since I don't have the 'and' operator yet :/

It doesn't run really fast, but that makes it a good optimization target maybe :)


sampsyo:
Looks super fun; nice work!! Could you add an entry to the docs with a one-line description of the benchmark?

benchmarks/bitwise-ops.bril
```diff
@@ -0,0 +1,95 @@
...
+ ret ans;
+}
+
+@main(a: int, b: int, c: int){
+  one: int = const 1;
+  zero: int = const 0; 
+
+  sel: int = sub c one;
+
+  ans: int = const zero;
```
sampsyo, line 75: Perhaps you mean `id zero`?

thomasyang18: Yup, meant that, nice catch. 

I added to the doc as well! Pretty new to pull requests, so is this all I gotta do?

sampsyo: Awesome! Yep, you got it.

Issue #139, opened 2022-02-01
Specify divide-by-zero semantics


sampsyo:
The Bril docs don't say anything about what divide-by-zero should do, either in the core instructions or in the floating point extension. Options include:

* Undefined behavior.
* Evaluate to zero (or NaN in the FP extension).
* Must crash with an exception.

Crashing with an exception is in some ways the most reasonable thing to do (and models many real languages), but it has two drawbacks:

* Exceptions are otherwise not a thing in Bril, so this would be a new concept.
* It makes basic manipulations more complex because you have to worry about potential side effects for `div` and `fdiv` instructions.

Of course, you can see the latter disadvantage as an advantage, pedagogically speaking… since it introduces an extra nuance, especially for speculative optimizations, that otherwise wouldn't exist in the language.

Pat-Lafon:
For the floating point extension, it seems reasonable to me that division-by-zero follows 64 bit IEEE 754 semantics since the documentation seemingly suggests that they can be treated as such. Both interpreters already follow these semantics. (So possible values are NaN, Infinity, and -Infinity for division by zero depending on the numerator).

For integers in the core instructions, I think it makes sense to return a division-by-zero error(current implementation-defined semantics are that both interpreters crash). Crashing is not helpful in terms of returning a readable error message.

> Exceptions are otherwise not a thing in Bril, so this would be a new concept.

This is referring only to core Bril right? Because they are a thing in the memory extension like in attempting to do an out-of-bounds access(The memory extension is probably also under-specified? Neither it's documentation nor it's blog post seem to mention that implementations will error when reading uninitialized data). So I don't think introducing a divide-by-zero error is too much of a surprise.

> It makes basic manipulations more complex because you have to worry about potential side effects for div and fdiv instructions.

If you want a fully compliant tool then they become more complex, however, one could always choose to take the `--ffast-math` approach and accept some level of incompatibility. In this way, one could always layer Undefined behavior on top of the current specification but one could not (with the current bril extensions) layer on an error for division-by-zero.

sampsyo:
> This is referring only to core Bril right? Because they are a thing in the memory extension like in attempting to do an out-of-bounds access(The memory extension is probably also under-specified? Neither it's documentation nor it's blog post seem to mention that implementations will error when reading uninitialized data).

[...]

PR #194, opened 2022-04-27
Make division by zero an error in brilirs

Pat-Lafon:
See #139

### DIFF
--- a/bril-ts/brili.ts
+++ b/bril-ts/brili.ts
@@ -462,7 +462,12 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
   }
 
   case "div": {
-    let val = getInt(instr, state.env, 0) / getInt(instr, state.env, 1);
+    let lhs = getInt(instr, state.env, 0);
+    let rhs = getInt(instr, state.env, 1);
+    if (rhs === BigInt(0)) {
+      throw error(`division by zero`);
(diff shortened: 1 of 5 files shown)


sampsyo:
Seems like a good idea to me, overall! I assume the test is failing because `brili` just returns `Infinity` or whatever? Maybe we should change it too to match…

Pat-Lafon:
The test currently fails because `brili` just executes the division by zero and doesn't gracefully handle the js `RangeError: Division by zero`. Based on the #139 discussion I see that this is allowable behavior, though probably not ideal for the reference interpreter.

Given this, should I move the test case to something like `test/brilirs-error/`?

> I assume the test is failing because brili just returns Infinity or whatever? Maybe we should change it too to match…

2ed5d33 Handled the mismatch on `Infinity`.

sampsyo:
Ah, neat! Here's another approach: I just fixed `brili` to exit gracefully with an error.

I didn't realize that, while `/` on plain old *numbers* in JavaScript returns a value without an error, doing it on `BigInt` objects raises an error. That's good! Anyway, we now detect this situation and raise our own error instead.

Pat-Lafon:
Works for me and looks like everything now passes.

sampsyo:
Awesome!

PR #196, opened 2022-05-05
make brench soft about floats

charles-rs:
minor change to brench to make it compare floats for approximate equivalence to deal with different ways of printing them

### DIFF
--- a/brench/brench.py
+++ b/brench/brench.py
@@ -46,15 +46,14 @@ def run_pipe(cmds, input, timeout):
         for proc in procs:
             proc.kill()
 
-
+ε = .0000000001
 
 def compare_output(o1, o2):
     def my_compare(x, y):
         try:
-            return abs(float(x)-float(y)) < .0000000001
+            return abs(float(x)-float(y)) < ε
         except ValueError:
             return x == y
-    
     return functools.reduce(lambda x,y : x and y, map(lambda x, y : my_compare, o1.split(), o2.split()))
 
 def run_bench(pipeline, fn, timeout):
@@ -129,8 +128,6 @@ def brench(config_path, files, jobs):
                 if first_out is None:
                     first_out = stdout
                 elif not compare_output(stdout, first_out) and not status:
-                    print(first_out)
-                    print(stdout)
                     status = 'incorrect'
 
                 # Extract the figure of merit.
--- a/brench/brench.py
+++ b/brench/brench.py
@@ -54,7 +54,7 @@ def compare_output(o1, o2):
             return abs(float(x)-float(y)) < ε
         except ValueError:
             return x == y
-    return functools.reduce(lambda x,y : x and y, map(lambda x, y : my_compare, o1.split(), o2.split()))
+    return all(my_compare(x, y) for x, y in zip(o1.split(), o2.split()))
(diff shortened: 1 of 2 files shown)


brench/brench.py
```diff
@@ -117,7 +128,9 @@ def brench(config_path, files, jobs):
                 # Check correctness.
                 if first_out is None:
                     first_out = stdout
-                elif stdout != first_out and not status:
+                elif not compare_output(stdout, first_out) and not status:
+                    print(first_out)
+                    print(stdout)
```
sampsyo, line 133: Looks like some stray debug printouts---probably delete these?

brench/brench.py
```diff
@@ -46,6 +47,16 @@ def run_pipe(cmds, input, timeout):
             proc.kill()
 
 
+
+def compare_output(o1, o2):
+    def my_compare(x, y):
+        try:
+            return abs(float(x)-float(y)) < .0000000001
```
sampsyo, line 54: Perhaps turn this threshold into a constant, `EPSILON`?

sampsyo (commented):
Awesome; thanks for adding this! Any chance you'd be interested in adding a note to the documentation for the tool so people know this is going on?

brench/brench.py
charles-rs: oops pushed the wrong version!

brench/brench.py
charles-rs: good call

charles-rs:
incoming documentation

docs/tools/brench.md
```diff
@@ -76,3 +76,8 @@ The latter is the value extracted from the run's standard output and standard er
 * `incorrect`: The output did not match the "golden" output (from the first run).
 * `timeout`: Execution took too long.
 * `missing`: The `extract` regex did not match in the final pipeline stage's standard output or standard error.
+
+A note on matching: floating point numbers are compared to be within an ε
+value. This is primarily motivated by different interpreters printing them
+differently, but it does mean that some unsound floating point algebraic
+simplifications may slip through
```
sampsyo, line 83: Suggested wording for clarity/completeness:

> To check that a run's output is "correct," Brench compares its standard output to that of the first run (`baseline` in the above example, but it's whichever run configuration comes first). The comparison is mostly an exact string match, but Brench also handles floating-point numbers differently by checking for numerical difference. This fuzziness deals with the many different ways one can print floats, but it also means that Brench could miss some unsound transformations that only slightly affect floating-point accuracy.

sampsyo (changes requested):
Thanks!! Just a couple of suggestions within.

Pat-Lafon:
Hi, I don't use brench myself but I saw this pr as I was making my own. Since you've already abstracted the magic number out to a constant, have you considered making it a user provided value with a default to zero? This would allow the tool to continue to work as expected for most users and then you can opt-in to a more generous epsilon value.

charles-rs:
@Pat-Lafon yeah that sounds like a good idea. done.

sampsyo (approved):
Changed to avoid using a mutable global variable. Other than that, looking great! Thanks for adding this!

PR #198, opened 2022-05-19
Cleaning up #197

Pat-Lafon:
The latest version of clippy introduced/improved some lints which are resolved in 7236a07. Unfortunately, it also seems like there is a false positive with one of the lints which has yet to be addressed. https://github.com/rust-lang/rust-clippy/issues/6902 

In my haste to fix a bug with positions, I accidentally flipped row and col for the initial position. Fixed in 8f0549c.


sampsyo:
Amazing; nice work!!!

PR #199, opened 2022-05-25
Start a Cranelift-based compiler

sampsyo:
As a way to procrastinate on grading, I tried writing a little Bril-to-native-code compiler using the fantastic [Cranelift](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/README.md). Learning more about Cranelift was fun!

Cranelift could hypothetically yield both an AOT and JIT compiler, and I've tried to keep the option open do both, although this version is just AOT. There is a little runtime library (in C) that you have to link in to get an executable. Cranelift's amd64 support seems a bit more mature support than its ARM support, so I've had a good ol' time running things through Rosetta 2 on my Apple Silicon machine. (Some scripts still have a hard-coded target to make cross-compilation a bit easier.) It would be cool to try investigating the blockers on the ARM path… none of them seem overwhelmingly complex.

This version only supports core Bril (no extensions). It agrees with the reference interpreter on all the non-memory, non-FP benchmarks, as shown by this command in the `benchmarks` directory:

    grep -L 'ptr\|float' *.bril | xargs turnt -c turnt_brilift.toml

I am resisting the temptation to add more language support and instead would like to force myself to make the testing and performance measurement story better before building more non-MVP features.

It also seems anecdotally to generate reasonably efficient code, but I have not done any systematic benchmarking yet, so I will resist the temptation to give any actual numbers, lest they turn out to be misleading.

I am still not very good at Rust, so maybe someone like @Pat-Lafon would like to critique my terrible code? No pressure, certainly!


docs/tools/brilift.md
```diff
@@ -0,0 +1,63 @@
...
+Build
+-----
+
+Brilift is a Rust project using the [bril-rs][] library.
+You can build it using [Cargo][]:
+
+
+    $ cd brilift
+    $ cargo build
+    $ cargo install  # If you want the executable on your $PATH.
```
Pat-Lafon, line 23: This is deprecated and wont work. Use `cargo install --path .` instead. Or `cargo install --path brilift` without the `cd brilift`. Or have something like `make install`. I don't think it's worth asking a potential user to run `cargo build`. If they want to run this code with out installing it, they would do this via `run` instead, otherwise if they want the executable, `install` is going to rebuild the project anyways in release mode.

docs/tools/brilift.md
```diff
@@ -0,0 +1,63 @@
...
+
+[bril-rs]: rust.md
+[cargo]: https://doc.rust-lang.org/cargo/
+
+Compile Stuff
+-------------
+
+Provide the `brilift` executable with a Bril JSON program:
+
+    $ bril2json < something.bril | brilift
```
Pat-Lafon, line 33: I tried to do this without looking at the code base(as a user) and it didn't seem to work. I tried to run `bril2json < ../benchmarks/ackermann.bril | brilift` and `bril2json < ../test/interp/tiny.bril | brilift` which both error. These programs both work when running with `brilift/run.sh` so I was surprised by this.

Pat-Lafon: The error I receive is

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unimplemented relocation addend Relocation { offset: 8, size: 26, kind: Relative, encoding: AArch64Call, symbol: SymbolId(5), addend: -4 }")', /Users/patricklafontaine/.cargo/registry/src/github.com-1ecc6299db9ec823/cranelift-object-0.83.0/src/backend.rs:500:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm curious why run.sh works without issue.

Pat-Lafon (changes requested):
I've briefly looked over this super cool pr. I don't think I have any major complaints but I didn't dive too deep into this. Most of my comments here relate to the documentation, which I attempted to follow and have requested changes. Since I had already made a local branch of this to look at the code with my editor, I figure I should just include any small code changes I thought would be useful in their own pr, #200. Feel free to take or ignore which ever commits you like.

Note that `../test/interp/main-bool-args.bril` fails which might be surprising.

Pat-Lafon:
I see that at one point you tried to make the runtime library in Rust but that didn't seem to work out. I am curious as to why? My guess would be that writing the c code for the parsing functions is easier.

Pat-Lafon:
I don't think I have much feedback in terms of the Rust code. From a high level it was all readable to me and made sense. I would like to think that between clippy and fmt, there is enough Rust tooling to make any Rust code seem normal.

I was a bit confused by the use of the `enum_map` crate which I hadn't seen before. (Not that this is a bad thing, I'm just unfamiliar). If possible, I'm curious to know why this was used.

docs/tools/brilift.md
sampsyo: Ah, thanks! Didn't realize this feature was removed. Updated!

docs/tools/brilift.md
sampsyo: Ah yes, this is the Apple Silicon trouble I alluded to in the original PR comment. I still want to work more on sorting this out, because I think it is either my fault or a simple fix on the Cranelift side—Cranelift has officially supported Apple Silicon for a while.

The reason `run.sh` works is because it specifies an x86_84 target instead of using the default architecture for the system, so everything is running through Rosetta 2.

sampsyo:
Thank you so much for taking a look!! This is all super helpful.

About writing the runtime library in Rust: I tried it two ways. The first was using a normal Rust package to create a statically linked library. For example, the `__bril_print_int` function just used `println!("{}", i)` or whatever. This worked OK and wasn't too hard, but it created a _very_ large library (tens of MB?) for what seemed like such simple functionality. Then I thought I'd try using libc's `printf` and such directly from Rust instead, and using `#[no_std]`, yielding code that looks like this:
https://github.com/sampsyo/bril/blob/568fab174049c85b178cbf67da16bb734e74d457/brilift/rt.rs#L3-L12

…which was definitely more minimal, but between the `unsafe` and the direct libc calls, it became more trouble than it was worth. I was basically writing C in Rust syntax, which wasn't very fun. It's not a terrible idea but it just seemed more practical to be honest that I was actually just writing C.

I found `enum_map` useful as an alternative to `HashMap` when the entire universe of keys was statically known. Prior to using the map, I was doing weird repetitive stuff for the runtime library, like this:
https://github.com/sampsyo/bril/blob/538695680120b2664a9eeeb8fcaff740fa2b10f3/brilift/src/main.rs#L15-L20

…which obviously caused a lot of duplication, but it satisfied my desire to avoid paying the cost of a whole hash table and real hashing and everything when I knew all the keys ahead of time anyway! Putting things in `enum_map` seemed like a reasonable compromise. It let me reduce lots of duplication by doing stuff like this:
https://github.com/sampsyo/bril/blob/5cf986e6fa1b9e8a741352419df78ea73c26795d/brilift/src/main.rs#L184-L195

…to do the same thing for every runtime function, but without needing some weird/expensive/verbose setup for iterating over all possible runtime functions. I hope that makes sense!

Pat-Lafon:
The use of c in the Runtime library makes sense.

At the risk of wading into something that you've thought a lot more about than I have, I have proposed #201 which tries to cut straight to the core code functionality needed here. I'm actually not too sure why you need to iterate over all possible runtime functions versus just declaring them on demand. The only worry I found was in avoiding declaring the same function twice, but that is manageable by just looking up the name first.

PR #200, opened 2022-05-26
Brilift

Pat-Lafon:
I would like to propose some code changes.

d7bf0ab - object files should probably be ignored
70bc6b3 - I would like to recommend running turnt via `make test` and `make benchmark` which is what I do for `brilirs`.
e69e3db - There was a comment about an unnecessary clone. I resolved this issue by adding a lifetime to the struct so it can contain references. Given the current structure of the code base, this works rather seamlessly but I haven't really thought about how it would affect future development. I'm not sure if you had already considered this design and found an issue with it. Feel free to ignore this commit if that is the case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ node_modules/
 hypothesis_*
 __pycache__
 .vscode
+
+**/*.o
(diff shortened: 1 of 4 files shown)


sampsyo:
All looks great! Thank you!!! The lifetime annotation for `CompileEnv` is an especially good idea; I may want to do that to some other members someday…

PR #201, opened 2022-05-26
Remove use of maps for refs

Pat-Lafon:
I tried to de-tangle the enum_map thing and came up with this. I moved the logic that was used in the map into a trait. The main logic difference is checking whether a name exists before trying to declare it. It seems to work based off of running the tests.

I think it's nicer because it doesn't have to look up any function names/enums to find the function reference and only declares the functions that it needs. 

Disclaimer, I had to rebase on the latest changes but I think I managed. I'm not thinking too hard about the overall structure of the code, so there may be good reasons not to structure the code this way(like moving Module into each of the structs.)


brilift/src/main.rs
```diff
@@ -7,14 +7,62 @@ use cranelift_codegen::settings::Configurable;
...
+    fn get_func_id<M: Module>(
+        &self,
+        module: &mut M,
+        call_conv: cranelift_codegen::isa::CallConv,
+    ) -> cranelift_module::FuncId {
+        match module.get_name(self.name()) {
+            Some(FuncOrDataId::Func(f)) => f,
+            Some(_) => panic!("Expected {} to be a function, not data", self.name()),
+            None => module
+                .declare_function(
```
sampsyo, line 33: On the other hand, it's possible that putting this `declare_function` call into a utility function that takes a `&mut Module` or whatever might be a good idea for concision…

brilift/src/main.rs
```diff
@@ -61,13 +113,13 @@ impl RTFunc {
...
 /// `main` function.
-#[derive(Debug, Enum)]
+#[derive(Debug)]
 enum RTSetupFunc {
     ParseInt,
     ParseBool,
 }
 
-impl RTSetupFunc {
+impl RTFuncUtils for RTSetupFunc {
```
sampsyo, line 122: Sharing infrastructure between the two categories of runtime functions may also be a good idea. The current separation conflates two things:

* some functions are only used in the C main wrapper
* those functions also happen to be the only ones whose signatures care what size pointers are on the target platform

Those are really not essentially conflated. It's possible that the best design would just combine them all into one enum…

brilift/src/main.rs
```diff
@@ -318,168 +353,175 @@ fn gen_binary(
 }
 
 /// An environment for translating Bril into CLIF.
-struct CompileEnv<'a> {
+struct CompileEnv<'a, M: Module> {
+    module: &'a mut M,
```
sampsyo, line 357: I'm actually a little wary of putting the `Module` into this environment. Part of the idea with separating `compile_inst` from the `Translator` structs was to _divorce_ it from the underlying module as much as possible. The reason is the same reason why Cranelift separates the core `cranelift_codegen` crate from `cranelift_module`: it's drawing a line between the intraprocedural "let's generate raw machine code sequences" functionality from the broader interprocedural context of stitching together multiple functions into a complete module. So despite the secondary goal of lazily only using the necessary runtime functions in a given Bril function, I would love to preserve this separation as much as possible…

brilift/src/main.rs
```diff
@@ -583,7 +606,7 @@ impl<M: Module> Translator<M> {
         }
 
         // Insert instructions.
-        compile_body(&func.instrs, &mut builder, &env);
+        env.compile_body(&func.instrs, &mut builder);
```
sampsyo, line 609: Moving `compile_body` and friends to an `impl` for `CompileEnv` is a good organizational idea!

sampsyo (changes requested):
Thank you for taking a look at all this!! There are some good ideas in here, and I've commented on a few. However, I'm a little uncertain about the core strategy for getting rid of `enum_map`, which involves directly looking up functions on the Cranelift `Module` object *on demand* instead of declaring them up front. I see the advantages in terms of laziness, but I mentioned inline a couple of concerns about the general design: (1) the intermingling of the ABI names as the identifiers for the runtime functions, and (2) the exposure of the `Module` to the instruction generation code, which in principle should only need `cranelift_codegen` to work. Let me know what you think!

brilift/src/main.rs
```diff
@@ -7,14 +7,62 @@ use cranelift_codegen::settings::Configurable;
...
+    ) -> ir::Signature;
+
+    fn name(&self) -> &'static str;
+
+    fn get_func_id<M: Module>(
+        &self,
+        module: &mut M,
+        call_conv: cranelift_codegen::isa::CallConv,
+    ) -> cranelift_module::FuncId {
+        match module.get_name(self.name()) {
```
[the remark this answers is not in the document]
Pat-Lafon: If one wants to continue using Cranelift's philosophy with https://docs.rs/cranelift-codegen/0.84.0/x86_64-apple-darwin/src/cranelift_codegen/ir/libcall.rs.html#141-166 then I guess it would be pretty straight forward to implement the behavior of EnumMap without the dependency. Maybe this would be the way to go given your feedback.

I'm not totally clear on this distinction that is being drawn with using the enum as a source of truth vs a string. Since all of this functionality is enclosed in an `impl` of the enum where the string name is never exposed to/provided by the user, then from an interface perspective, it doesn't look like there is a difference between what I've done and the current structure. I understand the implementation is different, but this is just a consequence of Cranelift under the hood mapping strings to functions. This is also something the current implementation has to do with `func_refs: HashMap<String, ir::FuncRef>`.

I came at this from a different view of efficiency which may in practice be less optimal(We know that only certain functions will be used, why are we paying the cost to declare functions that will never be used and then making additional allocations to store this information?).

brilift/src/main.rs
Pat-Lafon: I think I could pass a closure which hides away any underlying implementation of how from a given `RTFunc`/`String`, an `ir::FuncRef` is produced. This way you could do either method, using an EnumMap/HashMap or a function lookup, and from the `CompileEnv`'s perspective it is none the wiser. I think that encapsulates the separation you are looking for but maybe it's too much indirection.

sampsyo: FWIW, the alternative idea I was pondering for importing only the used functions was to do a second pass over the instructions… that is, an earlier pass could look for all functions referenced by any instruction and import those, all up front.

Pat-Lafon:
I haven't made much progress with this, and I haven't kept up with the latest changes to `brilift` so I will be closing this pr.

PR #206, opened 2022-06-02
Use turnt ancestors instead of -c hack

Pat-Lafon:


sampsyo (approved):
Wahoo; thanks for rearranging this!! I'm very glad this works. It restores the ability to run a single test by just doing `turnt <path_to_test>`, for instance, regardless of where you happen to be "standing." 🙏 🙌

PR #209, opened 2022-06-05
Docs cleanup

Pat-Lafon:
This pr is to follow up on https://github.com/sampsyo/bril/pull/197#pullrequestreview-978450064 in terms of cleaning up some of the documentation. I'm interested in any feedback for the import extension documentation, but would also be happy to do any `bril-rs`/`brilirs` related changes.


sampsyo (approved):
Awesome; all looks great! I have another pass over the docs in general on my to-do list; I'll send any relevant thoughts then.

PR #210, opened 2022-06-06
Tests: Use multiple Turnt environments

sampsyo:
I have added a new feature to Turnt to make multiple test configurations simpler to use. The rough idea is that, where we previously used several separate `turnt.toml` files, we can now put multiple commands into one config. See: cucapra/turnt#19 https://github.com/cucapra/turnt#multiple-environments

The main advantages are:

- can run multiple configurations with a single `turnt` invocation
- no need to duplicate test files
- convenient differential testing

This replaces our various `turnt_*.toml` files with these "environments" in the core tests and benchmarks. Only the "basic" tools are enabled by default; other tools need to be enabled explicitly with `-e brilirs` or similar. Seeing the `make test` targets in those relevant Makefiles will demonstrate the payoff. Paging @Pat-Lafon in case he'd like to take a look! The design isn't finalized or anything, so any feedback about how this currently works would be useful.

I also updated some of the example tests to use this feature too. With this, it's pretty easy to run the examples in multiple "modes" automatically, capturing the output for each.


Pat-Lafon:
Looks much cleaner.

In playing around with this, `turnt -e brilirs -e bril_rs $(TESTS)` works better than I expected.

I don't think I have much to add. The only thing I was thinking about were ways I would try and use multiple environments. This mostly ends up being trying to run all of the tests, across all of the tools, at once. Given that you have to supply all of the files you want to run for all environments, `turnt` will attempt to run some environments on tests that won't work. For example, running `test/interp` with both `brili` and `brilirs`, I would want to run `brili` on `test/interp/spec` but not with `brilirs`.

This is probably a non-issue but I was wondering if with this direction, maybe test files could acquired/filtered by `turnt` in a new way(maybe the environments in the toml file specify in some way what tests to run).

sampsyo:
Awesome; thanks for taking a look!

Yeah, some more complicated rules about when to run which tests could be interesting. I was pondering some kind of weird syntax like `turnt foo.bril:env1 bar.bril:env2` or similar to let you specify exactly which environment to run on which file. I don't love that specific UI, but maybe something like it could be viable in the future. There is also the potential for https://github.com/cucapra/turnt/issues/7, which would let you just type `turnt` to run a pre-configured list of tests, maybe.

PR #212, opened 2022-06-07
improve float handling/testing

Pat-Lafon:
As I was poking through #211 I realized that `bril2json`(rust) was pretty simplistic in it's handling of floats. The testing story for the floating point extension is a little minimal so this pr does both.

More parsing tests are added for defining floating point literals. I've correspondingly added on to the parsing regex to handle these cases. Neither implementation allows for defining a Nan/inf which is probably fine.

As I was looking at edge cases(mostly for comparisons), it seemed like `brili` was not handling `-0` correctly and was converting it to `0`. Not sure whether this is allowable behavior.


Pat-Lafon:
Apparently the `toString` method in js converts -0 into "0".

Pat-Lafon:
And it seems we have a discrepancy between the 2 bril2json implementations about whether `1e-6` should be outputted into JSON as `1e-6` or `1e-06`

sampsyo:
Well, that's all certainly exciting! @charles-rs noticed some related problems with perfect consistency in printing floats as resolved in https://github.com/sampsyo/bril/pull/196. Unfortunately, I'm not sure it's easy to get _perfect_ consistency between implementations on printing, but at least we can try to get closer! 

> Apparently the `toString` method in js converts -0 into "0".

That's really wild. Incredibly, it seems to require a hack to work around when stringifying JSON:
https://stackoverflow.com/a/19577454

Pat-Lafon:
I've attempted to patch `brili` to print out `-0` though I'm not that familiar with Javascript/Typescript. 

I've learned that there is a lot of nuance in how different languages print out floating point values with e-notation. I did come up with a python solution for removing the leading zero from `bril2json` which I've included below for future reference. But in evaluating the other tools, it seems like each language has it's own slight variations so I've instead commented out the tests using e-notation.

```python
## https://stackoverflow.com/questions/17503981/is-there-a-way-to-override-pythons-json-handler
## https://stackoverflow.com/questions/14863143/remove-leading-0-in-exponential-format-in-python
## https://github.com/toml-lang/toml/issues/356
class FloatEncoder(json.JSONEncoder):
    def __init__(self, **kwargs):
        super(FloatEncoder,self).__init__(**kwargs)

    # uses code from official python json.encoder module.
    # Same licence applies.
    def iterencode(self, o, _one_shot=False):
        """Encode the given object and yield each string
        representation as available.

        For example::

            for chunk in JSONEncoder().iterencode(bigobject):
                mysocket.write(chunk)
        """
        if self.check_circular:
            markers = {}
        else:
            markers = None
        if self.ensure_ascii:
            _encoder = json.encoder.encode_basestring_ascii
        else:
            _encoder = json.encoder.encode_basestring

        def floatstr(o,):
            # Check for specials. Note that this type of test is
            # processor and/or platform-specific, so do tests which
            # don't depend on the internals.
            return repr(o).replace("E-0", "E-").replace("e-0", "e-")

        _iterencode = json.encoder._make_iterencode(
                markers, self.default, _encoder, self.indent, floatstr,
                self.key_separator, self.item_separator, self.sort_keys,
                self.skipkeys, _one_shot)
        return _iterencode(o, 0)

```

sampsyo:
Nice!! This is a great workaround for `print` in the TypeScript interpreter at least.

That JSON representation thing is quite a bugaboo. I guess maybe the lesson here is that string comparison on the JSON documents is just not the right thing to do in testing—instead, we should somehow be comparing the parsed JSON values in AST-land??

PR #213, opened 2022-06-12
Simple implementation wall-clock benchmarking

sampsyo:
I wanted to set up a reasonably automated way to compare the execution time of Bril implementations on the benchmarks. This approach abuses a Turnt environment to invoke [Hyperfine][] and then aggregates the results. There is a little more usage documentation.

Tagging @Pat-Lafon because he might be interested in merging this with his existing [benchmark.sh](https://github.com/sampsyo/bril/blob/main/brilirs/benchmark.sh) and friends. Also including @charles-rs & @susan-garry because they might be interested in someday incorporating [their new bytecode interpreter](https://www.cs.cornell.edu/courses/cs6120/2022sp/blog/fast-brili/), described in https://github.com/sampsyo/cs6120/pull/337, into the benchmarking landscape.

Here are the harmonic mean speedups over brili for a recent run on havarti (an x86 server with dual 2.10 GHz Xeon Gold 6230s running Ubuntu 20.04 and Node v10.19.0), covering the core-only benchmarks:

```
brilirs: 63.81x
brilift-jit: 64.94x
brilift-aot: 186.56x
```

The setup can also draw a simple bar chart using [Vega-Lite][vl] (click to make legible):

plot

Basically, Brilirs is consistently a bit faster than the end-to-end execution time of Brilift in JIT mode, given the latter's extra startup cost for compilation, except for a few benchmarks that involve some nontrivial computation (e.g., `primes-between` and `pythagorean_triple`). Brilift in AOT mode doesn't get "charged" for compilation in this setup, so programs run about 3× faster. (But I wouldn't put too much stock in the exact numbers because so many benchmarks are very very small and run in basically no time.)

[hyperfine]: https://github.com/sharkdp/hyperfine
[vl]: https://vega.github.io/vega-lite/

### DIFF
--- a/benchmarks/.gitignore
+++ b/benchmarks/.gitignore
@@ -1,0 +1,3 @@
+*.bench.json
+bench.csv
+plot.svg
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -4,7 +4,7 @@ CORE_BENCH := $(shell grep -L 'ptr\|float' *.bril)
 bench:
 	turnt -e bench --save $(CORE_BENCH)
 clean:
-	rm -f *.bench.json
+	rm -f *.bench.json plot.svg bench.csv
 plot: plot.svg
 
 bench.csv: $(wildcard *.bench.json)
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -1,0 +1,14 @@
+CORE_BENCH := $(shell grep -L 'ptr\|float' *.bril)
+
+.PHONY: bench clean plot
+bench:
+	turnt -e bench --save $(CORE_BENCH)
+clean:
+	rm -f *.bench.json
+plot: plot.svg
+
+bench.csv: $(wildcard *.bench.json)
+	python3 summarize.py $^ > $@
+
+%.svg: %.vl.json bench.csv
+	npx -p vega -p vega-lite vl2svg $*.vl.json > $@
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -3,7 +3,7 @@ Bril Benchmarks
 
 This directory contains a suite of benchmarks in Bril contributed by the community. You can read [more details in the documentation][bench-docs].
 
-There is also some basic infrastructure here for benchmarking the wall-clock execution time of Bril implementations. Run the benchmarks comparing [the reference interpreter][brili], [brilirs][], and [brilift][] by typing:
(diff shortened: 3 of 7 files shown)


Pat-Lafon:
Cool, I am pretty willing to replace my hacky shell scripts with this. 

I figure I can throw a benchmarking toml in `brilirs/` that follows this same pattern. A bonus will be being able to use it to benchmark `brilirs` against previous versions of itself.

I will note that hyperfine has this warning about running code that takes less than 5ms to run. This is why I’ve mostly provided only longer running benchmarks in my comparisons between `brili` and `brilirs`(See `long_benchmark.sh`). The way around this is usually passing `—shell=none/-N` to hyperfine, which is fine for `brilirs` because it optionally supports taking file names as input instead of `stdin` via the `-f/—file` flag. I believe that the use of `stdin` for input requires hyperfine to use an intermediate shell. My understanding is that other interpreters do not have this functionality which is fine, but might make the smaller benchmark numbers reported unreliable. FYI, at the moment, I pretty consistently get the following results:

```
❯ hyperfine --shell=none --warmup 3 --export-json bench.json 'brilirs -f pow.json'
Benchmark 1: brilirs -f pow.json
  Time (mean ± σ):       2.2 ms ±   0.4 ms    [User: 0.6 ms, System: 0.4 ms]
  Range (min … max):     1.6 ms …   7.2 ms    1045 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```
```
❯ hyperfine --warmup 3 --export-json bench.json 'brilirs -f pow.json'
Benchmark 1: brilirs -f pow.json
  Time (mean ± σ):       1.3 ms ±   0.5 ms    [User: 0.6 ms, System: 0.5 ms]
  Range (min … max):     0.8 ms …   5.6 ms    876 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

It’s pretty surprising that the jit-ed code for `function_call.bril` is so much faster than `brilirs` and on par with the aot code. Seems like if you add a print statement it brings `brilift` within a factor of 2 from `brilirs`. So it’s probably doing something quite smart here since the currently `function_call.bril` has no side effects

sampsyo:
Ah, yes, I do think the main solution to this is to primarily look at the less-trivial benchmarks... or perhaps, someday, to provide multiple input "sizes" that lead to different running times for each benchmark.

Good point about the function call benchmark; maybe it's just getting optimized away entirely!

PR #215, opened 2022-06-21
brilirs deps + buffering

Pat-Lafon:
This pr handles the ever marching forward of dependency versions and switches to using `BufWriter` when writing to `stdout` which, as https://doc.rust-lang.org/std/io/struct.BufWriter.html describes, is much more efficient for many small calls to write.

Looking at this now, I realize that there is maybe one case where the `BufWriter` does not get flushed: when the Bril program raises an `InterpError` and then there is an error during the dropping of `BufWriter` where its contents are not flushed. I think this is both unlikely in practice and an allowable edge case.

### DIFF
--- a/bril-rs/bril2json/Cargo.toml
+++ b/bril-rs/bril2json/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["compiler", "bril", "parser", "data-structures", "language"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap         = { version = "3.1", features = ["derive"] }
+clap         = { version = "3.2", features = ["derive"] }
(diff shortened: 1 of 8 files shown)


sampsyo (approved):
Wahoo; all looks great!

Pat-Lafon:
I probably should have included some numbers to justify making a change for "performance reasons". Here is a silly benchmark of `function_call.bril` except it prints the argument it was passed at the end of the call.

| `brili` | old `brilirs` | new `brilirs` | `brilift -j` | `brilift` aot |
| -- | ---| --- | ----| --- | 
| 14.226 s ±  0.187 s | 2.128 s ±  0.008 s | 658.7 ms ±   3.3 ms | 1.772 s ±  0.022 s | 1.082 s ±  0.024 s |

PR #218, opened 2022-06-24
Fix #216

### LINKED ISSUE #216: Specify how `print` works for `float` values
Should other parts of floating point be specified as well, specifically how a compliant interpreter/runtime outputs `Nan/Infinity`?

_Originally posted by @Pat-Lafon in https://github.com/sampsyo/bril/issues/214#issuecomment-1161892561_

Also, we should specify that printing "normal" `float`s should always use 17 decimal digits, as discussed in https://github.com/sampsyo/bril/pull/214#issuecomment-1160827346.

Pat-Lafon:
This pr attempts to specify the output value of printing a float. A first pass of this is added to the documentation of the floating point extension. `brili` and `brilirs` have been modified to be compliant. The tests/benchmarks which output floats have also been updated.


sampsyo:
Awesome; looks perfect!! I just tweaked the docs a bit for precision (and moved the text to a new section just about printing).

I was going to suggest that we not include trailing zeroes, because C `printf` and Python string formatting make this easy with the `g` specifier:

```py
>>> '{:.17f}'.format(.5)
'0.50000000000000000'
>>> '{:.17g}'.format(.5)
'0.5'
```

…but it looks like Rust does not make this easy to do (see https://github.com/rust-lang/rust/issues/29472) so that's probably too much trouble.

Pat-Lafon:
Seems like this was somewhat discussed more recently in https://github.com/rust-lang/rfcs/pull/2729 but they didn't go so far as to fully implement `g` behavior. If this is something you would prefer, it naively seems like a relatively straight forward change for `brilirs`.

If it's just a matter of trailing zeros, then https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3ea18aef3548196ad33d8c6aa386364f seems reasonable.

If there is other behavior to be aware of, then there is atleast one crate that gives the desired behavior https://crates.io/crates/gpoint .

sampsyo:
Ah, good digging!! Honestly I think it's totally cool to use this zero-having output… at the very least it's simple to specify. We can always revisit if someone eventually gets super annoyed staring at `1.00000000000000000` instead of `1.0`. :smiley:

PR #219, opened 2022-06-25
Brilift: Conformant `float` printing

sampsyo:
Just makes Brilift conform to the new spec in #218 for printing all floats with 17 decimal digits of precision. Now all the float tests pass!

Hoping to tack CI stuff on to this PR…

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -36,8 +36,8 @@ jobs:
           - "cd bril-rs && make features"
           - "cd brilirs && make test TURNTARGS=-v"
           - "cd brilirs && make benchmark TURNTARGS=-v"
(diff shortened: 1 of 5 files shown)


Pat-Lafon:
As I'm passing through, I was curious about the failing test. Am I correct in assuming that `brilift` requires that `clang` is your c compiler? Where as `gcc` uses `--target`.

sampsyo:
That is a problem, although I didn't mean to include a target at all (and there's another problem somewhere else). (GCC is not a "cross-compiler by default" like Clang is, so you actually have to build a special GCC for compiling to non-native architectures.)

sampsyo:
Alright; I think everything's in order!

PR #220, opened 2022-06-26
Brilift: Support the memory extension

sampsyo:
This adds support for the memory (pointer) extension in Brilift. It works differently in the two modes:

* In AOT mode, it just calls libc's own `malloc` and `free`.
* In JIT mode, we use [Rust's `alloc`](https://doc.rust-lang.org/alloc/) to do something similar. Unlike `malloc`, however, `alloc::alloc` and `alloc::free` don't keep track of the sizes of allocations themselves, so we need to store those in the allocations themselves. This is not too bad.

A consequence of both techniques is that invalid memory accesses (e.g., out-of-bounds accesses or UAFs) are undefined behavior in both settings.

In JIT mode at least, it would be really cool to someday try using [Cranelift's sandboxed "heaps"](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/docs/ir.md#heaps) instead of actual Rust allocations. But I believe this would require including our own allocator somehow (which would operate within the regions of memory created by the JIT), so it may not be worth it.

Anyway, taken all together, this means that Brilift can run all the benchmarks.


sampsyo:
Looks like `adler32` is crashing in JIT mode in CI (although it does not do that on my machine). Fun!

Anyway, in the mean time, here's quick performance comparison on all 44 benchmarks, using the same (havarti) setup as in https://github.com/sampsyo/bril/pull/213#issue-1268552011. The summary is:

```
brilirs: 61.79x
brilift-jit: 60.71x
brilift-aot: 182.68x
```

In other words, basically the same averages as before.

plot

sampsyo:
Found the bug, although it's surprising it only arose for a single benchmark. I was accidentally freeing the memory for the generated machine code before calling the compiled function, so our `alloc` instructions could overwrite the program itself. `unsafe` is not a toy, kids!

PR #221, opened 2022-06-26
Optimizing Brilirs print

Pat-Lafon:
Working off of #215, I was kind of curious how much easy performance I could get out of my modified `function_call.bril` benchmark. Some of the easy wins are in avoiding a `Vec` allocation for single argument prints, avoiding `std::fmt/writeln!` as much as possible by just writing bytes to `stdout`, using `&str` instead of allocating a `String` when possible, and using https://crates.io/crates/itoa/1.0.2 for converting integers to `&str`.

| `brili` | old `brilirs` | new `brilirs` | `brilift -j` | `brilift` aot |
| -- | ---| --- | ----| --- | 
| 14.289 s ±  0.140 s | 674.1 ms ±  14.4 ms | 416.3 ms ±   8.1 ms | 1.776 s ±  0.026 s | 1.069 s ±  0.006 s |

I've also included the update I've made to `function_call.bril` as the second commit. By having a side-effect with each function call, it makes the benchmarking of interpreters more interesting since the function call can't necessarily be optimized away. However, it also makes an already long running benchmark take even longer and has a large amount of output which makes it great for seeing small interpreter improvements but maybe not ideal to have as a general benchmark(plus the output file is 5.6M lines long/11MB). Maybe this commit should not be included or moved somewhere else.

Disclaimer: This is a silly benchmark which is probably never representative of an actual Bril program.


Pat-Lafon:
One would similarly like like to use the https://docs.rs/ryu/latest/ryu/index.html crate for converting floats to strings(based on the PLDI paper of the same name). However, the implementation does not allow for specifying the necessary precision of 17 so I gave up trying to use that(though I got a great performance benefit when testing it). In doing so, I found a small bug in brilift and have included the fix + an improved test case.

test/interp/float/float.bril
```diff
@@ -1,4 +1,5 @@
-@main {
+# ARGS: 3 3.0
+@main(float_one : float, float_two : float) {
```
sampsyo, line 2: Maybe a test for `float`-typed `@main` arguments should be a separate test? (I'm imagining a future student coming along and wanting to implement `float` functionality one step at a time, and maybe a simpler example without `@main` arguments would be nice to have separate from this.)

brilift/src/main.rs
```diff
@@ -1013,7 +1013,7 @@ fn main() {
             .map(|(arg, val_str)| match arg.arg_type {
                 bril::Type::Int => bril::Literal::Int(val_str.parse().unwrap()),
                 bril::Type::Bool => bril::Literal::Bool(val_str == "true"),
-                bril::Type::Float => bril::Literal::Bool(val_str.parse().unwrap()),
+                bril::Type::Float => bril::Literal::Float(val_str.parse().unwrap()),
```
sampsyo, line 1016: Good catch; thanks!

sampsyo (commented):
Very cool; nice work on this!!

About the benchmark itself, I tend to agree that this one (even before the `print` was added) might be worth separating into a different benchmark "suite." Its purpose is clearly more about stress-testing implementations than providing a substrate for implementing typical 6120 optimizations (e.g., data flow analysis or DCE or whatever). What do you think about sequestering it in a separate directory, which may grow future stress-tests over time?

As a practical matter, the output file from the `print`-including version is now 11 MB. I'm not sure it's practical/useful to include this snapshot in the repository. Maybe this new/separate suite should not necessarily need to include output? (If possible, it would be great to expunge the large file from git history so future clones don't need to include it…)

test/interp/float/float.bril
Pat-Lafon: Sure

Pat-Lafon:
I'm not sure if I messed something up in terms of the benchmarking scripts used to make the plot. I am able create plots from the benchmarking times but the `brilift-aot` time is a lot slower than I would of expected(roughly a factor of 2 slower than `bril-jit`). I'm not sure why, maybe it's just something with the setup on my end, but this part in particular probably needs a second look over.

sampsyo (approved):
Looks great to me; thanks for all this!

My best guess for the weird benchmark results is that you're using Rosetta 2 (i.e., x86 binaries) for those AOT-compiled benchmarks? The JIT mode in Brilift supports ARM, but the AOT mode does not yet (because of some limitations in the Rust [object](https://crates.io/crates/object) crate).

PR #227, opened 2022-07-20
Rs2bril

Pat-Lafon:
I've been motivate to write `rs2bril` like the current `ts2bril`. I didn't quite realize until half way through that the stalled #107 also has an implementation of a `rs2bril` like tool. We approached the problem in a similar way so there was some repeated work though we are targeting different extensions of Bril so this pr has some set of new contributions. The implementation structure is similar enough that https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/bril-algebraic-data-types/ can be used as an overview.

#107's `rs2bril` targets core Bril + ADT's. `ts2bril` targets core Bril + Floats. This pr implements a `rs2bril` tool which targets core Bril + Floats + Memory. I've added a full list of limitations that I'm aware of to the README, I will highlight a few:
- Currently requires all variable declarations to be explicitly typed.
- `println!` is expected to be valid rust unlike in #107, however the format string is ignored during compilation and you can only pass variables as arguments.
- It always assumes valid Rust code(you can run all of your code as Rust first!), except in the case that you want to pass arguments to main. In this case, just declare function arguments to main like any other function(which makes it not valid Rust).
- Memory is implemented as Rust Arrays/Slices. This means that they are statically sized with integer literals and must be initialized on creation. I've also specialized `drop` to compile to `free`. Programmers need to manually manage their memory(by dropping arrays). This could be a use case for https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/asmm/ .

It can provide source positions via the `-p` flag though many are probably off/missing.

`example.rs` has some example syntax that is supported. I've added a set of tests which are equivalent to the `ts2bril` tests plus more complicated benchmarks like `ackermann.rs`, `riemann.rs` and `cholesky.rs`(which is probably the best representation of what `rs2bril` can do).


sampsyo:
Wow; super cool that this works!! I really like the idea to support passing around slices as the way to get Bril pointers. It's a nice compromise between Rustiness and practicality for mapping to the Bril memory extension.

I'm sure there's plenty more to do, but do you think this is ready for "v1" to be merged into the monorepo?

Pat-Lafon:
> I'm sure there's plenty more to do, but do you think this is ready for "v1" to be merged into the monorepo?

I think I've largely accomplished what I set out to do(be able to compile and run one of the non-trivial memory-based benchmarks in both the Rust Playground and as Bril). It has some kind of documentation which includes it's limitations(which can double as future work) along with a passing test suite/lint checks. I don't have any further commits planned for this pr beyond any changes you may ask for or bug fixes that come up.

sampsyo:
Awesome! Super great work on this; it looks extremely cool.

PR #228, opened 2022-07-20
pgo for brilirs

Pat-Lafon:
Profile-Guided Optimization kind of seems like a modest amount of free performance(~5%). Cargo/Rust make using LLVM's support for this rather easy so I've included a shell script + make commands to compile `brilirs` with this optimization.


sampsyo (approved):
Looks good from here! Cool that this is not too tricky to get going.

PR #229, opened 2022-07-20
Actual BrildErrors in Brild

Pat-Lafon:
Adds some structured errors to Brild via `thiserror`. There are still a couple `expect`/`panic` but I believe these would be triggered because of a bug and not because of user input.


sampsyo (approved):
Great. I didn't know about the `thiserror` crate; this looks like a convenient way to go about this kind of setup.

PR #232, opened 2022-07-26
Richer Source Positions

Pat-Lafon:
Now that there are a couple of front-end compilers, specifically looking at #227 and #197, I want to tie things together into a nice bow by extending #161 with (optional) richer source position information as discussed in part in #191.

To support something span-like, this pr adds an optional `pos_end` field to describe the end line/column of the position.

To provide source code for the error message, this pr adds an optional `src` field which is an absolute path to the original file that this position points to. Note that the `src` field does not need to be the same across the entire Bril program.

Each is currently described as their own field to be compatible with the original specification, but I would probably like to combine them into one object if that's fine. Something like 

```JSON
{
  ...
  "pos": {
    "pos_start" : {"row": <int>, "col": <int>},
    "pos_end" : {"row": <int>, "col": <int>}?,
    "src" : "<string>"?,
  }
}
```

### DIFF
--- a/docs/lang/syntax.md
+++ b/docs/lang/syntax.md
@@ -124,11 +124,13 @@ It also has the optional `args`, `funcs`, and `labels` fields.
 Source Positions
 ----------------
 
-Any syntax object may optionally have a `pos` field to reflect a source position:
+Any syntax object may optionally have position fields to reflect a source position:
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Nice!! Pretty rad that this extra information can be done in a clean backwards-compatible way. Looks perfect to me!

PR #233, opened 2022-10-03
bril2json -p -p and Rust/Clap updates

Pat-Lafon:
- All the rust tools now use Clap v4.0
- Rust 1.64 Clippy lint updates
- Link fix in docs
- Minor bug in brilck where it didn't check if there were actually instructions in the function(func.instrs could be `undefined`) which triggered on `test/check/badcall.bril` for `@nothing`. Not sure why this wasn't caught before.
- bril-rs adds support for #232
- (rust) bril2json now takes multiple `-p` flags where just one `-p` is still compatible with python versions and `-p -p` additionally adds the end source positions.
- rs2bril now supports end source positions via `proc_macro2::Span`(best effort) and also adds the `--file/-f` flag.
- brilirs now has rough support for end source positions.

I've manually tested the additional support for richer source positions but like the rest of the error-handling support is not rigorously tested.

### DIFF
--- a/brilck.ts
+++ b/brilck.ts
@@ -281,14 +281,14 @@ const INSTR_CHECKS: {[key: string]: CheckFunc} = {
     }
     return;
   },
-  
+
   phi: (env, instr) => {
     let args = instr.args ?? [];
     if (!('type' in instr)) {
       err(`phi needs a result type`, instr.pos);
       return;
     }
-    
+
     // Construct a signature with uniform argument types.
     let argTypes: bril.Type[] = [];
     for (let i = 0; i < args.length; ++i) {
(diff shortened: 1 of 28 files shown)


brilck.ts
```diff
@@ -353,25 +353,27 @@ function checkFunc(funcs: FuncEnv, func: bril.Function) {
...
   // Gather up all the types of the local variables and all the label names.
-  for (let instr of func.instrs) {
-    if ('dest' in instr) {
-      addType(vars, instr.dest, instr.type, instr.pos);
-    } else if ('label' in instr) {
-      if (labels.has(instr.label)) {
-        err(`multiply defined label .${instr.label}`, instr.pos);
-      } else {
-        labels.add(instr.label);
+  if (func.instrs){
```
sampsyo, line 356: Good catch!

sampsyo (approved):
Nice! This all looks great!!

This is probably for a subsequent PR, but maybe we should just make the enhanced positions thing the default everywhere… that is, we could make the Python-based parser behave the same way, and alleviate the need for the `-p -p` mode in the Rust-based parser. Does that seem reasonable to you?

Pat-Lafon:
> This is probably for a subsequent PR, but maybe we should just make the enhanced positions thing the default everywhere… that is, we could make the Python-based parser behave the same way, and alleviate the need for the `-p -p` mode in the Rust-based parser. Does that seem reasonable to you?

Seems reasonable to me.

A separate question, maybe source positions should be spans by default(instead of starting positions by default) if all of the tools that support source positions provide starting and ending positions. This would mean making the `pos_end` field not optional(though providing positions as a whole is still optional).

sampsyo:
Yeah, seems absolutely reasonable to me.

PR #234, opened 2022-11-02
Brilirs Simple bench

Pat-Lafon:
- Replaces the current shell scripts with the technique of #213 for `brilirs` benchmarking.
- Adjusts `summarize.py` to not error if not all modes are used when benchmarking.
- Adds logic to detect arm64 platforms and set `TARGET` for brilift. 

Note, probably cleaner to review after #233 is merged(assuming it doesn't have any blockers) as this is based on that branch to avoid merge conflicts.


benchmarks/.gitignore
```diff
@@ -1,4 +1,3 @@
-**/*.bench.json
-bench.json
+*.json
```
sampsyo, line 1: Not sure about this—some JSON files are actually meant to be checked in. For example, the outputs of the parser tests:
https://github.com/sampsyo/bril/blob/main/test/parse/add.json

benchmarks/summarize.py
```diff
@@ -57,6 +57,9 @@ def summarize(bench_files):
             'speedup': speedup,
         })
 
+    ## Filter out modes which don't have data, and thus no speedups
+    speedups = {k: v for k, v in speedups.items() if v != []}
```
sampsyo, line 61:
```suggestion
    # Filter out modes which don't have data, and thus no speedups.
    speedups = {k: v for k, v in speedups.items() if v}
```

Empty lists are falsy, so no need for `!= []`.

sampsyo (commented):
Great!! Here are two very tiny initial comments.

benchmarks/.gitignore
sampsyo: Oops, thanks, didn't realize this was in a subdirectory!

Pat-Lafon:
History should be cleaner now. With the new release of 1.65 today, I've addressed any new clippy warnings.

sampsyo (approved):
Wahoo; nice work on this!!

PR #235, opened 2022-11-28
Brillvm: Bril-rs to LLVM

Pat-Lafon:
This is a semi-interesting Thanksgiving break project inspired by `bril-llvm` #138 and `brilift` #199 to compile the Rust representation of Bril from `bril-rs` to LLVM IR.

This is different from `bril-llvm` in two ways:
- It uses a Rust wrapper around LLVM called Inkwell https://github.com/TheDan64/inkwell to construct the `*.ll` file using the LLVM IRBuilder instead of concatenating strings together which should be more extensible.
- It avoids performing an SSA conversion on the Bril program by allocating all of the variables on the stack at the start of the function. You can then load variables from the stack, perform an operation, and then store the result back to the stack with temporary variables(which I think is what Clang does?).

Currently, this tool supports core Bril and the floating point extension ~except for passing command line arguments~.

~I have semi-arbitrarily chosen to support LLVM 14. It is both the latest supported version of LLVM provided by Inkwell and the current version used by my Clang installation.~ I have chosen to support LLVM 15 as it is now the latest supported version of LLVM by Inkwell and the current version used by my rust compiler(>1.64)

At ~1k lines of Rust(but 800 lines for manually building each instruction), I don't think anything would be too surprising. 
- The tool operates in two iterations: the first iteration creates the initial versions of all of the functions and populates the first basic block with stack allocations for all of the variables, the second iteration then builds all of the instructions for each function.
- ~The main function is converted to have a return type of int and always returns 0 to indicate success.~ The main function is renamed to "_main" and is called by the real main function after command line arguments has been parsed. Still always returns 0.
- Boolean values are a little tricky because LLVM stores them as integers(this requires holding onto typing information to remember which values are bools and a cast to i8 for `_bril_print_bool`)
- Extra logic is added to ensure all basic blocks have a terminating instruction. This means that unconditional jumps are added instead of falling through and any dead code after a terminating instruction but before a new label is skipped.

Some things that could be improved:
- ~Add support for the memory/ssa extensions~
- ~Correctly handle passing arguments to main~ (cli arguments are now supported)
- Documentation
- The print functions are written externally and included as LLVM bytecode at compile time. ~Ideally, this would be a rust file which I would compile to llvm with `rustc`. However, the current version of `rustc` uses LLVM 15 and I had enough difficulty getting this to work on an older version of `rustc` that I've punted on this. Instead I compile `brilift/rt.c` to a `rt.bc` file using Clang~ In borrowing from `brilift`, I have my own `brillvm/bin/rt.rs` file which is compiled to a `rt.bc`. This prints via the `libc_print` crate and handles the parsing of c strings to their corresponding values.

Additional Edits:
- There is a little bit of unsafe code used in the LLVM GEP instruction and in parsing raw c strings into their safe rust `CStr` counter parts. Both of these are involved in parsing command line arguments. Bad things will probably happen if the correct arguments are not provided at runtime.
- `brillvm` now passes the `core`/`float` test and benchmark suites.
- `brillvm` is not responsible for producing fast LLVM code, thus I have not attempted any benchmarks and have only correctness tests. Interested users can use the appropriate LLVM tools to get performance numbers.
- ~`brillvm` has not been added to the github workflow because of the dependence on a LLVM installation at both compile and runtime. Therefore the current "passing" of the pr is not meaningful.~
- Overall, this project is a little rough but I greatly enjoyed learning more about LLVM.


sampsyo:
So sorry for somehow missing this the first time around!! It's seriously cool, and a great demonstration of LLVM codegen. Really fun project!

> It uses a Rust wrapper around LLVM called Inkwell https://github.com/TheDan64/inkwell to construct the `*.ll` file using the LLVM IRBuilder instead of concatenating strings together which should be more extensible.

I've been admiring Inkwell from afar for some time; cool to see it in action. I hope others take inspiration if they work in Rust.

> It avoids performing an SSA conversion on the Bril program by allocating all of the variables on the stack at the start of the function. You can then load variables from the stack, perform an operation, and then store the result back to the stack with temporary variables(which I think is what Clang does?).

Yes indeed, this is "the way" in the LLVM frontend ecosystem! Generate a bunch of `alloca`s for mutable variables and then rely on `mem2reg` to optimize them away.

> Extra logic is added to ensure all basic blocks have a terminating instruction. This means that unconditional jumps are added instead of falling through and any dead code after a terminating instruction but before a new label is skipped.

Similarly "the way"! Awesome that it wasn't too tricky to ensure terminators. (It's a restriction I considered enforcing in Bril too back in the day, but I think the current state is a little truer to its loosey-goosey nature.)

In general, looks awesome!! Feel free to merge whenever you feel like it, even if it's not 100% complete.

Pat-Lafon:
Awesome! I don't think I have any improvements to this code in mind at the moment but can make adjustments as needed.

It covers the main extensions of float, mem, and ssa; passing all test cases and benchmarks except cholesky. A few of the floating point values for that benchmark are off by a small epsilon (i.e. 250.01098702698604370 instead of 250.01098702698598686) which I didn't figure out at the time.

It mimics `brilift` in having both AOT and JIT modes though with no optimizations.

It has minimal documentation and gets checked by clippy.

It's pinned to LLVM 15 and relies on a corresponding rust compiler using the same llvm version to create the runtime library.

sampsyo:
Sounds great!! Seems like an exciting starting point for plenty of other stuff too.

Pretty weird about the minor float accuracy differences! I wonder if that's an LLVM thing or some kind of x86 vagary…

PR #236, opened 2022-12-13
Char type

tmaradc:
Char type implementação:
- Definição do tipo, utilizado com `const`;
- Operações: `ceq`, `clt`, `cle`, `cgt`, `cge`;
- Pode ser utilizado como parâmetro para a função `print`;
- Pode ser utilizado como tipo para vetores (ptr);

Ainda falta:
- Testes com o `turnt`;
- Possibilidade de o utilizar com a ferramenta [Bril Text Format](https://capra.cs.cornell.edu/bril/tools/text.html#bril-text-format)


sampsyo:
Seems very cool! I'd be happy to merge this if you're interested in contributing it. :smiley:

tmaradc:
> Seems very cool! I'd be happy to merge this if you're interested in contributing it. smiley


Thank you! I accidentally opened the PR draft, I still want to make some tests, but I'd love to contribute! 

I'm also thinking about a string type, but first I want to introduce the char type. What do you think? 

Soon I intend to open a PR!

sampsyo:
Sounds wonderful! As does starting with `char` and then moving on to a variable-length string type.

brili.ts
```diff
@@ -830,6 +883,10 @@ function parseMainArguments(expected: bril.Argument[], args: string[]) : Env {
         let b: boolean = parseBool(args[i]);
         newEnv.set(expected[i].name, b as Value);
         break;
+      case "char":
+        let c: number = parseNumber(args[i]);
```
Pat-Lafon, line 887: parseChar no?

Pat-Lafon:
I was curious if you had a specification for what is a valid char? I think that has implications for what encoding a compliant implementation would need to use(ASCII/UTF-8/UTF-16 which is what Javascript uses apparently) but I'm not super familiar.

Maybe related to the first question, how similar are chars to ints? What would be the difference between an operation like `ceq(c1, c2)` versus `eq(char2int(c1), char2int(c2))` if you had `char2int` and `int2char` conversion functions?

tmaradc:
> I was curious if you had a specification for what is a valid char? I think that has implications for what encoding a compliant implementation would need to use(ASCII/UTF-8/UTF-16 which is what Javascript uses apparently) but I'm not super familiar.

Good point, I still don't have it... I'm also not super familiar with that. But it is something I'm going to define before making a real PR.

> Maybe related to the first question, how similar are chars to ints? What would be the difference between an operation like `ceq(c1, c2)` versus `eq(char2int(c1), char2int(c2))` if you had `char2int` and `int2char` conversion functions?

The idea is to make something similar to C chars, so they would be similar. These two operations would return the same value. But I'm still thinking about it, how do you think it should be?

I'm really at the beginning of the idea, this PR (that was supposed to be a draft to my forked repository) was me trying to get familiar with the code.

brili.ts
tmaradc: Yes! I will change it.

Pat-Lafon:
I was just wondering if there were interesting tradeoffs between have a list of builtin char operations(like `ceq`) versus converting a char into an int and then relying on Bril int operations.

For example, operations like checking that a char is an alphanumeric character, checking character equality, or doing a lower to upper case conversion could be implemented in Bril code with just additional `char2int`/`int2char` functions.

However, the specification of an `int2char` function might be kinda weird(because not all valid ints are valid chars). Also, the above makes a couple of assumptions about what a character is(which I think is that they are ascii characters? I'm not that sure). Note that ascii can't represent a character like `ã` so that may be undesirable.

tmaradc:
I see... maybe it is indeed better to convert to int for the operations. I will think more about it.

PR #237, opened 2022-12-16
Address Rust 1.66 clippy lints

Pat-Lafon:
As described in title. I'm particularly happy about the fixing of a false positive in the `use_self` lint.

In doing this, I ran into an error on the typescript side: [error: TS2339 [ERROR]: Property 'pos' does not exist on type 'never'.](https://github.com/sampsyo/bril/commit/8b1c4754e16a7cf0966f5e6f1974b436763b26ea) My understanding is that typescript is saying that this branch is impossible end up in. I'm not sure how to handle this, so after trying a few permutations, I decided to trust the typescript typechecker and just removed the branch.

### DIFF
--- a/brilck.ts
+++ b/brilck.ts
@@ -318,10 +318,6 @@ function checkOp(env: Env, instr: bril.Operation) {
 }
 
 function checkConst(instr: bril.Constant) {
-  if (!('type' in instr)) {
-    err(`const missing type`, instr.pos);
-    return;
-  }
   if (typeof instr.type !== 'string') {
     err(`const of non-primitive type ${typeFmt(instr.type)}`, instr.pos);
     return;
--- a/brilck.ts
+++ b/brilck.ts
@@ -319,7 +319,7 @@ function checkOp(env: Env, instr: bril.Operation) {
 
 function checkConst(instr: bril.Constant) {
   if (!('type' in instr)) {
-    err(`const missing type`, instr!.pos);
+    err(`const missing type`, instr.pos);
     return;
   }
   if (typeof instr.type !== 'string') {
--- a/brilck.ts
+++ b/brilck.ts
@@ -318,6 +318,11 @@ function checkOp(env: Env, instr: bril.Operation) {
 }
 
 function checkConst(instr: bril.Constant) {
+  if (!(instr as any)) {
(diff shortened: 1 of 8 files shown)


brilck.ts
```diff
@@ -318,10 +318,6 @@ function checkOp(env: Env, instr: bril.Operation) {
 }
 
 function checkConst(instr: bril.Constant) {
-  if (!('type' in instr)) {
```
sampsyo, line 321: Maybe we can just do `(instr as any)`? This is a "defensive programming" kind of trick, where this check is meant to trigger in case the JSON data doesn't match the expected type from our declarations.

sampsyo (commented):
Hell yes; looks awesome! Here's a thought on that TS thing.

I notice Clippy has some nits about this doc comment:

```
/// For example "ptr<bool>" => Parameterized("ptr", Box::new(Primitive("bool")))
```

It seems to think `<bool>` is an HTML tag. We can probably correct that by putting it in code delimiters (backticks)?

Pat-Lafon:
> I notice Clippy has some nits about this doc comment:
> 
> ```
> /// For example "ptr<bool>" => Parameterized("ptr", Box::new(Primitive("bool")))
> ```
> 
> It seems to think `<bool>` is an HTML tag. We can probably correct that by putting it in code delimiters (backticks)?

I've updated the workflow to catch these issues and have made this fix.

brilck.ts
Pat-Lafon: I've made this change.

The need for this trick kind of surprises me because it sounds like the type is `bril.Constant` when it really should be `bril.Constant | undefined`?(Or the same corresponding argument for any fields that might not exist)

sampsyo: Yeah, it's true. The thing is, we want these type definitions to reflect what a "real" Bril program looks like. But at some point, we are doing `JSON.parse` on a Bril program and giving it the type `bril.Program` or whatever—without validating it. Validating the entire program up front would be a huge pain, so this type-cast is a little bit of a lie. But then we can back it up later with some of these ad-hoc checks that second-guess the IR type.

sampsyo (approved):
Awesome!! Thanks again!

PR #238, opened 2023-01-13
Tweaks

Pat-Lafon:
Just some small tweaks
- Up crate versions for clap and regex
- Feature gate brilirs shell completions. I think these are really cool, but they are a bit finicky(especially being located in the build script) so I've hidden this behind a feature flag for those who want it.
- Derive `PartialEq` and `Eq` when possible on the rest of `bril-rs` structs


sampsyo:
Super nice; thanks!! I agree with feature-gating the completions; like you said, they are nice to have but weirdly complicated for what they are (nice to have).

PR #240, opened 2023-03-05
[bril-rs] Derive Hash for Literal if float is not enabled

lanza:


sampsyo:
With @Pat-Lafon's 👍, looks good!!

PR #242, opened 2023-03-09
small clippy lint for Rust 1.68

Pat-Lafon:
Only a small change for the new Rust release.

I'll note that there is an additional future-incompat warning that we hit because of the current version of lalrpop. https://github.com/rust-lang/rust/pull/103418/ https://github.com/lalrpop/lalrpop/issues/701 

In one sense, it's kind of a non-issue because it doesn't hold up the ci and I think this can only become a hard error over an edition change. On the other hand, it's a little disconcerting if lalrpop isn't getting much attention from the maintainers these days.

I would expect that a (new?) maintainer will arise who can cut a new lalrpop release which will resolve this issue. Otherwise this could motivate a switch in parsers.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -43,7 +43,9 @@ jobs:
       - uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
-          components: cargo-hack
+
+      - name: Install cargo-hack for feature-flag checking
+        run: cargo install cargo-hack
 
       - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


Pat-Lafon:
Speaking of being unmaintained, apparently `actions-rs` is suffering from this issue. Here there seems to be a good amount of movement to `dtolnay/rust-toolchain` like with https://github.com/rust-lang/regex/pull/883/commits/bbb9daca7faebc385467615dca44cce6be182d3d so I feel reasonably confident in switching over.

sampsyo:
Very nice; many thanks, as always!

It's really too bad about lalrpop; there are just not that many good options for solid parser generators in the ecosystem (is pest the only viable alternative? There's Chumsky but that's a very different style).

And thanks for switching to [rust-toolchain](https://github.com/dtolnay/rust-toolchain); looks great!

Pat-Lafon:
I've explored using https://github.com/hydro-project/rust-sitter to leverage the tree-sitter ecosystem but it doesn't have the same kind of maturity yet(being built into the proc-macro attribute system of Rust does have a certain appeal/benefits).

PR #244, opened 2023-04-20
[ts2bril] Recognize top level code below function definitions as main function

agentcooper:
Previously top-level instructions were incorrectly added to function above, resulting in an empty main function.

```ts
function getThree(): bigint {
    return 3n;
}

var v = getThree();
console.log(v);
```

```
@main {
}
@getThree: int {
  v0: int = const 3;
  ret v0;
  v: int = call @getThree;
  v: int = id v;
  v1: int = id v;
  print v1;
  v2: int = const 0;
}
```

### DIFF
--- a/bril-ts/builder.ts
+++ b/bril-ts/builder.ts
@@ -134,4 +134,8 @@ export class Builder {
     this.nextFresh += 1;
     return out;
   }
+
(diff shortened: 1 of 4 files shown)


sampsyo (approved):
Aha; this makes sense!! Nice and clean solution here.

PR #245, opened 2023-04-20
Bump Rust dependencies

Pat-Lafon:
No interesting changes from the new Rust release (1.69).

I am taking this opportunity to update dependencies. The main upgrade to motivate this pr is moving to Syn 2.0. Also of note is that lalrpop made a point release to fix some of the previous issues. I look forward to their new maintainers + Niko publishing a 1.0 release.

I have also tweaked the workflow with respect to the ci flags.


sampsyo (approved):
Awesome work, as always!! And great news about lalrpop; I’m glad it’s getting some much-needed attention.

PR #247, opened 2023-05-01
[ts2bril] Support for pointers and memory operations

agentcooper:
This PR introduces support for built-in functions to manage heap memory. Functions are modelled after the [memory extension operations](https://capra.cs.cornell.edu/bril/lang/memory.html).

Example code that is now supported by `ts2bril`:

```ts
import * as mem from "../../bril-ts/mem";

let size = 100n;
let arr = mem.alloc<bigint>(size);

for (let i = 0n; i < size; i = i + 1n) {
    mem.store(mem.ptradd(arr, i), i); // arr[i] = i
}

let sum = 0n;
for (let i = 0n; i < size; i = i + 1n) {
    sum = sum + mem.load(mem.ptradd(arr, i)); // sum += arr[i]
}
console.log(sum);

mem.free(arr);
```

I had to update the TypeScript compiler version because the old version was missing `checker.getTypeArguments`.

### DIFF
--- a/bril-ts/mem.d.ts
+++ b/bril-ts/mem.d.ts
@@ -1,0 +1,9 @@
+/**
+ * https://capra.cs.cornell.edu/bril/lang/memory.html
+ */
+export interface Pointer<T> {}
+export function alloc<T>(size: bigint): Pointer<T>
+export function store<T>(pointer: Pointer<T>, value: T)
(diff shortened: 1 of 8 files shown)


sampsyo (approved):
This is very nifty! I like the idea to use a set of intrinsics to directly reflect the underlying instructions. Seems very useful for constructing benchmarks.

PR #249, opened 2023-05-14
Bug fix in tdce tests.

zfnmxt:
Looks like `double-pass.out` actually needed that second pass. 🙃


sampsyo:
Hi! Can you please explain more about why you want to make this change? This test is meant to exercise the single-pass version of dead code elimination, as indicated by this directive:

```
# ARGS: tdcep
```

sampsyo:
No problem!

PR #253, opened 2023-05-25
Char type extension

tmaradc:
Adds the char type extension, with support in `brili`, `bril-txt` and `brilck`.
Updated the TypeScript Library.


bril-txt/briltxt.py
```diff
@@ -210,6 +222,12 @@ def float(self, items):
     def nullptr(self, items):
         return 0
 
+    def char(self, items):
+        value = str(items[0])[1:-1] # Strip `'`.
+        if value in control_chars.keys():
```
sampsyo, line 227:
```suggestion
        if value in control_chars:
```

Just doing `key in dict` suffices to check for key existence.

bril-txt/briltxt.py
```diff
@@ -210,6 +222,12 @@ def float(self, items):
     def nullptr(self, items):
         return 0
 
+    def char(self, items):
+        value = str(items[0])[1:-1] # Strip `'`.
```
sampsyo, line 226:
```suggestion
        value = str(items[0])[1:-1]  # Strip `'`.
```

Just a tiny PEP8 style thing.

docs/lang/char.md
```diff
@@ -0,0 +1,35 @@
+Character
+==============
```
sampsyo, line 2:
```suggestion
Character
=========
```

docs/lang/char.md
```diff
@@ -0,0 +1,35 @@
...
+
+Types
+-----
+
+The character extension adds one new base type:
+
+    "char"
+
+Characters are [UTF-16][] code units.
```
sampsyo, line 11: This is totally fine and no need to change it for now, but someday we might want to instead specify that this is any single Unicode character instead. Tying stuff to UTF-16 is a little annoying, even though it's what we'd inherit from JavaScript.

docs/lang/char.md
sampsyo, line 1: I think this may need to be added to the table of contents, in `SUMMARY.md`.

sampsyo (commented):
Wow!!! This is super cool and very complete. Really nice work on the overall design, the opcodes, etc.; it seems like a model extension. I have a couple of very low-level code suggestions, but then this seems good to merge. Very cool!!

docs/lang/char.md
tmaradc: done!

docs/lang/char.md
tmaradc: Ok

sampsyo:
Thanks for the updates!! The CI uncovered a couple of type errors in some test files. Checking this file:
https://github.com/sampsyo/bril/blob/72174be07959a30a3dd8ded357b6a5a10fbca957/test/interp/mixed/store-char.bril#L27-L29

Yields these errors:

```
$ bril2json -p < test/interp/mixed/store-char.bril | brilck
27:3: hello has type ptr<char>, but arg 0 for load should have type ptr<bool>
28:3: cpos1 has type ptr<char>, but arg 0 for load should have type ptr<bool>
29:3: cpos2 has type ptr<char>, but arg 0 for load should have type ptr<bool>
```

It seems like some of those `: bool` annotations may need to be changed to `char`.

tmaradc:
> Thanks for the updates!! The CI uncovered a couple of type errors in some test files. Checking this file:
> 
> https://github.com/sampsyo/bril/blob/72174be07959a30a3dd8ded357b6a5a10fbca957/test/interp/mixed/store-char.bril#L27-L29
> 
> Yields these errors:
> 
> ```
> $ bril2json -p < test/interp/mixed/store-char.bril | brilck
> 27:3: hello has type ptr<char>, but arg 0 for load should have type ptr<bool>
> 28:3: cpos1 has type ptr<char>, but arg 0 for load should have type ptr<bool>
> 29:3: cpos2 has type ptr<char>, but arg 0 for load should have type ptr<bool>
> ```
> 
> It seems like some of those `: bool` annotations may need to be changed to `char`.

Oh no. I fixed it and updated the checks!

sampsyo:
Wahoo; looks like it's green now!

PR #255, opened 2023-06-07
Char issues

tmaradc:
Some issues were pointed out in this [PR](https://github.com/sampsyo/bril/pull/254).

I addressed some:

- Fixed the `bril-txt` tool to "round trip" files with control characters;
- Disallow lone "surrogate code points" for the `int2char` function. 

About the control characters: 
- In order for `bril2json < ../test/interp/char/char_args.bril | brili '\n' e y` to not errors out, `brili` should 'convert' the control character as `bril-txt` does, because it is read as two code points, and not a single newline character.
- If `brili` is not dealing with the scaped characters, I think printing values like `\u000b` is a consequence, I'm not sure how to deal with this.


brili.ts
```diff
@@ -762,7 +762,7 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
 
   case "int2char": {
     let i = getInt(instr, state.env, 0)
-    if(i > 65535 || i < 0) {
+    if(i > 65535 || i < 0 || (55295 < i && i < 57344)) {
```
tmaradc, line 765: > We could create a bug for supporting beyond-BMP characters in the JavaScript 

Does it mean making this range larger, even though chars are UTF-16 code points? Allow integers from 65535 to 1114111? (all the unicode scalar values)

bril-txt/briltxt.py
```diff
@@ -254,6 +254,9 @@ def type_to_str(type):
 
 def value_to_str(type, value):
     if not isinstance(type, dict) and type.lower() == "char":
+        control_chars_reverse = {y: x for x, y in control_chars.items()}
+        if ord(value) in control_chars_reverse:
+            value = control_chars_reverse[ord(value)]
```
sampsyo, line 259: Excellent!

brili.ts
sampsyo: Good question. Yes, I suppose the focus of that proposed issue would be removing the `i > 65535` part of this condition: supporting larger values that map to a single Unicode character but multiple JavaScript characters (i.e., JavaScript strings with `len()` greater than 1).

For those values, it looks like [`fromCharCode(n)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode) won't work by itself, but it might be as easy as switching to [`fromCodePoint`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint), which would be awesome and convenient if it works!!

sampsyo (commented):
Awesome; this looks great!! I think this is pretty much good to merge even without further `brili` changes; this seems wonderful on its own.

brili.ts
tmaradc: I think I got it...

tmaradc:
I added some changes, now the `char` type is a unicode character
WDYT? @sampsyo @Pat-Lafon

brili.ts
```diff
@@ -440,7 +440,7 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
       else
         value = BigInt(Math.floor(instr.value))
     } else if (typeof instr.value === "string") {
-      if(instr.value.length > 1) throw error(`char must have one character`);
+      if([...instr.value].length !== 1) throw error(`char must have one character`);
```
tmaradc, line 443: The spread operator is unicode-aware so surrogate pairs won't be a problem.

brili.ts
```diff
@@ -754,18 +754,18 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
...
   case "int2char": {
-    let i = getInt(instr, state.env, 0)
-    if(i > 65535 || i < 0) {
+    let i = getInt(instr, state.env, 0);
+    if(i > 1114111 || i < 0) {
       throw error(`value ${i} cannot be converted to char`);
-    }
-    let val = String.fromCharCode(Number(i));
+    } 
+    let val = String.fromCodePoint(Number(i));
```
tmaradc, line 768: I maintained the range because `fromCodePoint` throws if the value is not a valid code point

brili.ts
```diff
@@ -754,18 +754,18 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
...
+    let val = BigInt.asIntN(64, BigInt(code as number));
     state.env.set(instr.dest, val);
     return NEXT;
   }
 
   case "int2char": {
-    let i = getInt(instr, state.env, 0)
-    if(i > 65535 || i < 0) {
+    let i = getInt(instr, state.env, 0);
+    if(i > 1114111 || i < 0) {
```
Pat-Lafon, line 765: I just want to double check if this range is sufficient to rule out "lone surrogates". I'm just guessing based purely off of comparing against how a rust `char` is defined so I'm not 100%. Note how there is a gap in Rust's char range where the surrogate values are: https://doc.rust-lang.org/std/primitive.char.html#validity

tmaradc: I was searching more, and since [fromCodePoint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#compared_to_fromcharcode) is according to the UTF-32 code unit, I guess an integer in the surrogate range will return an [ill-formed](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#page=54) string. It is still a valid code point though, that's why the `fromCodePoint` won't throw with the surrogate range.

I think we can prevent that.

sampsyo (approved):
Awesome!! This looks great! I'm an especially big fan of these test output files:
https://github.com/sampsyo/bril/blob/ec0f81b56b624beb29d2661b327d223b1de0a1af/test/interp/char/char.out#L1
https://github.com/sampsyo/bril/blob/ec0f81b56b624beb29d2661b327d223b1de0a1af/test/interp/char/char_args.out#L1

😸

Pat-Lafon:
@sampsyo FYI, there seems to be an issue in the github action for deploying the documentation.

sampsyo:
Indeed—the host that serves as the "bastion server" to deploy the updated documentation has been down for a while, and being physically distant this year, I haven't been able to fix it. I hope I can figure out an alternative solution sometime!

PR #256, opened 2023-06-10
Bug Fix: Variable Overwrite Causing Algorithm Error

yslking:
I have identified and fixed a bug in `examples/lvn.py` that occurs when a variable from another basic block is overwritten within the current basic block and then used as a parameter again.

To resolve the bug, I have introduced a modification. Before the variable is first written to, I copy its value to a new variable `lvn.i.xxx`.

Example:

```bril
@main {
  x: int = const 1;
.lb:
  y: int = id x;
  x: int = add x x;
  print y;
}
```

After running the transformation using `python examples/lvn.py -p -c -f`, the code is transformed as follows:

```bril
@main {
  x: int = const 1;
.lb:
  y: int = id x;
  x: int = add x x;
  print x;
}
```

In this case, `x` is overwritten before it is used again in `print x`, causing the algorithm to fail.

After applying the bug fix and running the code again, the transformation produces the following result:

```bril
@main {
  x: int = const 1;
.lb:
  y: int = id x;
  lvn.i.0 = id x;
  x: int = add x x;
  print lvn.i.0;
}
```

Now, the code is functioning correctly.

### DIFF
--- a/examples/lvn.py
+++ b/examples/lvn.py
@@ -96,12 +96,22 @@ def lvn_block(block, lookup, canonicalize, fold):
     # Track constant values for values assigned with `const`.
     num2const = {}
 
+    external_vars = set()
+
+    new_instrs = []
+
     # Initialize the table with numbers for input variables. These
     # variables are their own canonical source.
     for var in read_first(block):
         num = var2num.add(var)
         num2var[num] = var
+        external_vars.add(var)
 
+    # If a var is defined outsides of the block, and the instruction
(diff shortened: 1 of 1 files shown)


sampsyo:
AWESOME; very nice work tracking down a very tricky bug in the example implementation. I ended up concocting a slightly different fix in #257; that one has the slight advantage of not requiring a generated `id` instruction and a new name for this example test case. I added your test there.

Thanks again for pointing this out!

PR #258, opened 2023-08-24
Small batch of Bril/Rust fixes

Pat-Lafon:
- A little bit of cleanup of now unnecessary allowed lints/workflow components
- Bumping to Inkwells support of llvm 16
- Fixed the small floating point issue of brillvm by specifically asking for an execution engine without optimizations
- Today's release of Rust 1.72's clippy lints


sampsyo (approved):
Yay; nice work as always!! I went over everything and it looks A-OK. 👍

PR #259, opened 2023-08-27
Mandelbrot and Factorial Benchmarks

stephenverderame:
Bril benchmarks for an ASCII mandelbrot set and a simple factorial for lesson 1 tasks.


sampsyo:
Very nice! The Mandelbrot one is particularly fun to look at. :smiley: Did you use the TypeScript frontend (or some other frontend) to generate it?

stephenverderame:
Thanks, I used the typescript frontend

PR #260, opened 2023-08-29
Add cordic benchmark

matth2k:


benchmarks/float/cordic.out
```diff
@@ -0,0 +1 @@
+0.86369602123419631
```
sampsyo, line 1: Just confirming with a reference implementation:

```py
>>> import math
>>> math.sin(1.0472)
0.8660266281835431
```

Accurate to two decimal places! Not too shabby.

sampsyo (approved):
Wow; super cool benchmark!! This algorithm is fascinating.

Scutua (approved):

PR #261, opened 2023-08-29
[Benchmark] conjugate gradients

willwng:
A bril benchmark for using conjugate gradients to solve `Ax=b`. Currently the file takes in `n`, which is the size of the matrix and `A` is diagonal with incrementing values. However, the code implements dense matrix multiplication and can support any arbitrary `A` (it is somewhat tricky to randomly generate spd matrices).

fyi: the vim syntax highlighting also sometimes stopped working when functions had a specified return type


sampsyo:
Really cool; nice work!

There is a minor conflict on the docs page; could you please resolve that so we can merge?

sampsyo:
Thanks!! Now that the CI has run, it has also revealed that the benchmark has a few type errors. (You can [run the type checker yourself](https://capra.cs.cornell.edu/bril/tools/brilck.html) if you want.) In particular, it looks like some places there is a `ptr<int>` where it should probably be a `ptr<float>`.

willwng:
> Thanks!! Now that the CI has run, it has also revealed that the benchmark has a few type errors. (You can [run the type checker yourself](https://capra.cs.cornell.edu/bril/tools/brilck.html) if you want.) In particular, it looks like some places there is a `ptr<int>` where it should probably be a `ptr<float>`.

Oops, didn't realize there was a type checker! The github workflow integration is really neat. I think I fixed the type errors and added an input for the size of the matrix.

sampsyo:
Looks great; thanks!

PR #262, opened 2023-08-29
reverse_int

MelindaFang-code:
reverse the integer
e.g. 123 -> 321


benchmarks/core/reverse.bril
```diff
@@ -0,0 +1,31 @@
+
+@main{
+  input: int = const 123;
```
sampsyo, line 3: This is OK, but to make the benchmark a little more reusable, you might consider making the input a parameter. Here's an example:
https://github.com/sampsyo/bril/blob/a7a0759b20e724260825fc9c6ae7273a09c0bafd/benchmarks/core/armstrong.bril#L1-L2

That is, you can make `input` a parameter to `@main` and then give it a value like `# ARGS: 123` for benchmark automation purposes.

sampsyo (commented):
Super cool and compact idea! Here is one minor suggestion. And it would also be great to include #263 to this PR.

sampsyo:
Hmm… it looks like this may have erroneously reformatted the whole `bench.md` file?

MelindaFang-code:
ooops it seems that it autoformats when I save the file. I have pushed a fix. @sampsyo

sampsyo:
Looks good; thanks!

PR #265, opened 2023-08-30
Add Hanoi benchmark

bcarlet:


benchmarks/core/hanoi.bril
```diff
@@ -0,0 +1,26 @@
...
+  call @hanoi above spare dst src;
+.else:
+  ret;
+}
+
+@main {
+  disk: int = const 2;
+  src: int = const 0;
+  dst: int = const 2;
+  spare: int = const 1;
```
sampsyo, line 24: FWIW, if you're interested, you could pass these as parameters from the test harness instead of hard-coding them in `@main`. That would (hypothetically) make it easier to run the benchmark on different inputs. Here's an example:
https://github.com/sampsyo/bril/blob/a7a0759b20e724260825fc9c6ae7273a09c0bafd/benchmarks/core/armstrong.bril#L1-L2

sampsyo (commented):
Looks good! If you're interested, I have one minor/unimportant suggestion about leaving the inputs as parameters.

benchmarks/core/hanoi.bril
bcarlet: Updated to take the number of disks as a parameter. The other constants are more or less baked in to the design of the game, so I left those hard coded.

sampsyo:
Looks perfect; thanks.

PR #267, opened 2023-08-30
Added palindrome benchmark

alifarahbakhsh:
Benchmark for 6120'23 task 2


sampsyo:
Hello! Your test output file isn't matching because it doesn't have a trailing newline. You can make sure the file is exactly right by using `turnt --save` to produce it.

sampsyo:
Getting closer! It's now missing the profiler output. Please use `turnt --save` and remember to commit the `.prof` file.

sampsyo:
Everything looks green now; thanks!!

PR #268, opened 2023-08-30
Add Euler's totient function benchmark

obhalerao:
Add a benchmark that computes Euler's totient function on an input n.

Overall a fun little task, though I initially tried to write the code in TypeScript and use ts2bril, but I found the experience so frustrating and the error messages so cryptic that I gave up and just wrote it directly in bril.


sampsyo:
So far so good; thanks!! Can you please add an entry to the docs?

sampsyo:
Excellent; thanks!!

PR #271, opened 2023-08-31
Add primitive root benchmark

SanjitBasker:
Add a benchmark that finds a primitive root modulo p for some prime p.


sampsyo:
Wow; looks awesome!

PR #275, opened 2023-08-31
Add benchmark for computing euclidean norm

evanmwilliams:


sampsyo:
Sounds cool! The next steps here are to add an entry to the docs and include the test outputs (you can use `turnt --save` to produce these).

evanmwilliams:
I see - added all these in!

sampsyo:
Looks good; thanks!!

PR #279, opened 2023-08-31
Add Bit Shift Benchmark

collinzrj:


sampsyo:
Looks great!

benchmarks/core/bitshift.bril
```diff
@@ -0,0 +1,81 @@
+# ARGS: 3 5 10000 4
+@pow(x: int, n: int): int {
+  v1: int = id n;
+  v2: int = const 1;
+  v3: bool = eq v1 v2;
+  br v3 .then.0 .else.0;
+.then.0:
+  v4: int = id x;
+  ret v4;
+  jmp .endif.0;
```
Pat-Lafon, line 10: Would you consider removing this dead jump and associated label? While this doesn't affect an interpreter, this trips up both `brilift` and `brillvm` as during compilation they get tricked in generating empty returns for the final label.

This could be fixed in both tools, but it's kind of a weird edge case the more I looked into it. I lean more towards this should be avoided in benchmark files than this is worth the added complexity in the above tools.

Is this a bug in the typescript compiler?

Let me know what you think.

sampsyo: Good question about whether this is something the TypeScript compiler should avoid… maybe a good step for investigating this would be to try to reproduce it with a minimal TS program that produces this weird CFG pattern?

Scutua (approved):

PR #280, opened 2023-08-31
Add IsDecreasing benchmark

NgaiJustin:


benchmarks/core/is-decreasing.bril
```diff
@@ -0,0 +1,47 @@
+# Input: an array of length 6
```
sampsyo, line 1: Looks like this may be out of date? The input appears to be an integer, not an array.

sampsyo (commented):
Overall looking good! I found one possibly out of date comment, and it would be awesome if you could help resolve the merge conflict.

benchmarks/core/is-decreasing.bril
NgaiJustin: Oops good catch. Just fixed this!

sampsyo:
Excellent!

PR #281, opened 2023-08-31
Adding CSR Sparse Matrix Dense Vector Multiplication Benchmark

yxd97:
This benchmark generates a CSR sparse matrix and a dense vector with a linear feedback shift register RNG, multiplies them, and prints the matrix, intput, and output. I have verified it against the scipy implementation of matrix-vector multiplication.


sampsyo (approved):
Wow! This is really ambitious. Nice work getting the end-to-end thing to work, including the random matrix generator!!

benchmarks/mem/csrmv.bril
```diff
@@ -0,0 +1,268 @@
...
+    call @print_arr res rows;
+
+    free csr_rowptr;
+    free csr_colidx;
+    free csr_values;
+    free vec;
+    free res;
+
+    good: int = const 0;
+    ret good;
```
Pat-Lafon, line 267: Would you consider removing the return value of main? While I guess this hasn't been specified, afaik there isn't any implementation that will make use of this value. Most will silently ignore it like `brili` does.

`brilirs` rejects this program(I thought because of #91 but git blame predates that) which I think is fair?

Let me know what you think.

yxd97: I agree. Indeed there is no specifiction saying `main` shoud return a value to be used. 
Shall I make a new PR for this?

sampsyo: Good idea! Yes please, a new PR would be great. (I suppose we should document this, and probably add a check in `brilck`…)

yxd97: Just created #293

PR #282, opened 2023-08-31
Quicksort with Hoare Partitioning and Median of Three Pivot Selection

zachary-kent:
Benchmarks an implementation of quicksort with Hoare partitioning and median of three pivot selection. We then generate several arrays with pseudorandom elements, sort them, and check that they are indeed in nondecreasing order by 
printing out "true" if the array is sorted correctly.

It seems @AliceSzzze and I had the same idea, but it would still be fun to see what optimizations have different impacts on different partitioning schemes!


AliceSzzze:
Great minds think alike! Also median of 3 is definitely a better heuristic than just picking the last element lol

sampsyo:
Super cool! It's interesting that this benchmark is "self-checking" (instead of just outputting the sorted random array). Works fine! But I wonder if, someday, we should imagine having some other kind of non-snapshot test harness for benchmarks like this…

PR #284, opened 2023-08-31
Adding nilakantha pi approximation benchmark

JohnDRubio:
Added Bril implementation of [nilkantha's infinite series](https://www.youtube.com/watch?v=3U7Hfz37GqA&ab_channel=GaryExplains) to float testbench.


benchmarks/float/nilakantha.out
sampsyo, line 1: I think this may be the wrong output—it looks like the output for your block extractor/CFG builder. Maybe try `turnt --save nilakantha.bril` once again?

sampsyo (changes requested):
Super rad! This is a fun algorithm. It looks like some of the expected-output files may need updating.

PR #286, opened 2023-08-31
Adding sum-check benchmark

Arthur-Chang016:
Compute the sum of [1, n] by both loop and formula and compare them to see if the result is the same.


sampsyo:
Looks perfectly reasonable; thanks!

PR #287, opened 2023-09-01
birthday paradox benchmark

jpramos-me:


jpramos-me:
@sampsyo Not sure how I missed the `.prof` file, but here’s a new commit with it included!

sampsyo:
No worries! Thanks for catching that!! Looks good.

PR #288, opened 2023-09-01
Quickselect benchmark

ryanwmao:
Worked on this benchmark in person with Albert Xiao


sampsyo:
Very nice!! It's a cool twist on just doing plain ol' sorting.

PR #290, opened 2023-09-06
Brench: make glob recursive so it can match **

keikun555:
Sometimes we want to run brench on test files and benchmarks. This allows that


sampsyo:
Great idea; thanks!

PR #292, opened 2023-09-08
Specify and check that the `@main` function can't have a return type

sampsyo:
As discussed in https://github.com/sampsyo/bril/pull/281#discussion_r1319112205 and previously in https://github.com/sampsyo/bril/issues/91, it seems like `@main` shouldn't return anything. (Maybe we will change this in the future—e.g., maybe it should be able to return an OS status code—but we can relax this requirement if we ever want to do that.)

This PR adds this requirement to the docs and adds the obvious check to `brilck`. The CI will fail on this PR until the benchmark from #281 is fixed to obey the new rule.

### DIFF
--- a/brilck.ts
+++ b/brilck.ts
@@ -394,6 +394,14 @@ function checkProg(prog: bril.Program) {
   // Check each function.
   for (let func of prog.functions) {
     checkFunc(funcEnv, func);
+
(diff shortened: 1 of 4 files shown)


sampsyo:
All set now, following #293!

PR #293, opened 2023-09-08
remove the return value of main in csrmv

yxd97:
Removing the return value of the main function of csrmv, according to the discussion in #281.


sampsyo:
Wonderful; thanks!

PR #294, opened 2023-09-09
Miscellaneous fixes

Pat-Lafon:
- Update actions/checkout to v4
- Fixed an edge case in rust `bril2json` positions for the first row
- Javascript has some pretty loose parsing rules, I modified `brili` to be a little more strict on the cli input it expects (related to mod_inv)
- Modified the `ts2bril` tool to not unconditionally output the final endif label and corresponding jump when it isn't needed(related to bitshift and fact)
- Changed brilirs to check for main function return types on `--check`

The typescript stuff could use a double check, I'm not super familiar with this language. I also don't use `ts2bril` to know all of the edge cases.

### DIFF
--- a/bril-ts/builder.ts
+++ b/bril-ts/builder.ts
@@ -67,7 +67,7 @@ export class Builder {
       return this.buildEffect("call", args, [func], undefined);
     }
   }
-  
+
   /**
    * Build a constant instruction. As above, the destination name is optional.
    */
@@ -122,6 +122,32 @@ export class Builder {
       throw "cannot build instruction/label without a function";
     }
     this.curFunction.instrs.push(instr);
+  }
+
+  /**
+   * Checks whether the last emitted instruction in the current function is the specified op code.
+   * Useful for checking for terminating instructions.
+   */
+  isLastEmittedEffectOp(op: bril.EffectOpCode): boolean {
+    if (!this.curFunction) {
+      return false
+    }
+
+    if (!this.curFunction.instrs) {
+      return false
+    }
+
+    if (this.curFunction.instrs.length === 0) {
+      return false
+    }
+
+    const last_instr : bril.Instruction | bril.Label = this.curFunction.instrs[this.curFunction.instrs.length - 1];
+
+    if ('label' in last_instr) {
+      return false
+    }
+
+    return last_instr.op === op
   }
 
   /**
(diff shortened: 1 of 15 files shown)


bril-ts/builder.ts
```diff
@@ -124,6 +124,32 @@ export class Builder {
     this.curFunction.instrs.push(instr);
   }
 
+  /**
+   * Checks whether the last emitted instruction in the current function is the specified op code.
+   * Useful for checking for terminating instructions.
+   */
+  isLastEmittedEffectOp(op: bril.EffectOpCode): boolean {
```
sampsyo, line 131: Seems like useful behavior! But the name is a little odd, in two ways:

* `Emitted`: Maybe it would be a little more straightforward to just say it's the last op currently in the function? It seems less material that it's the last thing we emitted.
* `Effect`: I think this would look for both effect and value ops, right?

Another alternative, FWIW, would be to just get the last (non-label) instruction. So the return type would be `bril.Instruction?`, and callers could get this behavior by doing stuff like `lastInstr()?.op === "add"`.

brili.ts
```diff
@@ -873,7 +873,12 @@ function parseBool(s: string): boolean {
 
 function parseNumber(s: string): number {
   let f = parseFloat(s);
-  if (!isNaN(f)) {
+  // parseFloat and Number have subtly different behaviors for parsing strings
+    // parseFloat ignores all random garbage after any valid number
+    // Number accepts empty/whitespace only strings and rejects numbers with seperators
+  // Use both and only accept the intersection of the results?
+  let f2 = Number(s);
+  if (!isNaN(f) && f === f2) {
```
sampsyo, line 881: Weird but true! It'll do.

ts2bril.ts
```diff
@@ -215,12 +215,18 @@ function emitBril(prog: ts.Node, checker: ts.TypeChecker): bril.Program {
         // Statement chunks.
         builder.buildLabel(thenLab);
         emit(if_.thenStatement);
-        builder.buildEffect("jmp", [], undefined, [endLab]);
+        const then_branch_terminated = builder.isLastEmittedEffectOp("ret");
```
sampsyo, line 218: A little tricky, but makes sense!

sampsyo (commented):
A bunch of good stuff here! I have one thought about the builder library utility… let me know what you think.

bril-ts/builder.ts
Pat-Lafon: Sure, works for me.

I'll just note that I don't think the the `?` sugar works on return types? https://github.com/microsoft/TypeScript/issues/51332 Maybe I missed something obvious.

Instead I'm using `undefined` which the rest of the code base seems to do.

sampsyo (approved):
This looks perfect; thanks!!

PR #298, opened 2023-10-06
Rust 1.73 updates

Pat-Lafon:
See commit messages

### DIFF
--- a/bril-rs/bril2json/Cargo.toml
+++ b/bril-rs/bril2json/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["compiler", "bril", "parser", "data-structures", "language"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap         = { version = "4.3", features = ["derive"] }
+clap         = { version = "4.4", features = ["derive"] }
 lalrpop-util = { version = "0.20", features = ["lexer"] }
 regex = "1.9"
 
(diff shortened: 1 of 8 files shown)


bril-rs/bril2json/Cargo.toml
```diff
@@ -14,7 +14,7 @@ keywords = ["compiler", "bril", "parser", "data-structures", "language"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap         = { version = "4.3", features = ["derive"] }
+clap         = { version = "4.4", features = ["derive"] }
```
sampsyo, line 17: It's obviously not important now, but it occurs to me that we could consider setting up a Cargo [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) to centralize dependency versions like this.

sampsyo (approved):
All looks great; thanks a million as always!

bril-rs/bril2json/Cargo.toml
Pat-Lafon: I can, but there are two usability issues as far as I can tell.

One I knew about previously is that if you run a cargo command at the workspace level(compared to the crate level) then it will try to union all the possible feature flags of each dependency together. This is incompatible with some of the projects that only expect a subset of the `bril-rs` flags(a `bril*` tool that doesn't expect `bril-rs` to be built with positions for instance).

One minor thing I just figured out is that there are other settings that previously local would be forced to be workspace wide. Specifically for `brillvm` it is important to have `panic = "abort" lto = true` set for building/linking the `no_std` rt library. With a workspace, this gets forced upon all of the projects which isn't ideal(but not breaking/work around-able in this case I think). https://github.com/rust-lang/cargo/issues/8264#issuecomment-635424772

sampsyo: Makes sense! That first argument seems like sufficient reason not to do this (since `bril-rs` is meant to be a very configurable crate).

PR #299, opened 2023-10-07
brilift: Bump Cranelift dependency to v0.100.0

sampsyo:
Just a regular update to the latest version of Cranelift! This time, there was no API breakage. And in fact, it seems like, somewhere since the last time I updated things, Cranelift added support for AOT compilation (and `.o` file emission) on macOS-arm64. So we no longer need hacks to force things to run in Rosetta2 on Apple Silicon; I have taken these hacks out.

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -14,4 +14,4 @@ bench.csv: $(wildcard **/*.bench.json)
 	python3 summarize.py $^ > $@
 
 %.svg: %.vl.json bench.csv
-	npx -p vega -p vega-lite vl2svg $*.vl.json > $@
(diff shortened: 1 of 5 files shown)


sampsyo:
Just thought I'd include an updated, equally unreadable bar chart for the latest version running on my M1 Max MacBook Pro:

plot

The average speedups are:

```
brili: 1.00x
brilirs: 12.50x
brilift-jit: 10.87x
brilift-aot: 29.28x
```

I still don't think these are all that meaningful because so many of our benchmarks are extremely short-running (skewing things heavily in favor of AOT, for example), but that's a problem for another day.

PR #301, opened 2023-10-18
Add Hash, PartialOrd, Ord to Argument struct

hatoo:
Hi, It's very convenient if `Argument` has Hash, PartialOrd, Ord to me. (I want to put `Argument` as a key of some container)

Thanks!


sampsyo:
Based on the :+1: from @Pat-Lafon, I concur! Thanks!

PR #302, opened 2023-10-24
add dead-branch benchmark

2b3o4o:
Adds a new benchmark, dead-branch. This can be used to test a compiler's ability to detect and prune a dead branch, where a br instruction always evaluates to false and is therefore unnecessary.


sampsyo:
OK, sounds good! Most of our benchmarks are more like "real computations" than tests for optimizers, but not all. I'm good with merging this, but maybe someday we want to separate out these two categories of benchmarks…

2b3o4o:
Makes sense - thanks for the feedback :)

PR #303, opened 2023-10-29
Add return code support to brili

keikun555:
Brili doesn't set the return code of what the main function of bril returns. This fixes that.

### DIFF
--- a/brili.ts
+++ b/brili.ts
@@ -941,7 +941,7 @@ function evalProg(prog: bril.Program) {
     curlabel: null,
     specparent: null,
   }
-  evalFunc(main, state);
+  let retVal = evalFunc(main, state);
 
   if (!heap.isEmpty()) {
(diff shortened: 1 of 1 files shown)


sampsyo:
This is an interesting proposal! I'm certainly open to the overall idea. But if we do add it, there are a couple of things we'll need to do:

1. Decide whether we want the return value to be optional or required. I think the only workable option is to allow `@main` to return either nothing or `int`, i.e., yes, it's optional (and returning nothing produces an exit status of 0).
2. We'll want to extend various other parts of this repo to reflect the change:
  * Add a description of the functionality to the docs.
  * Add tests.
  * Add to the reference type checker and ensure that no existing tests/benchmark disobey the rules (i.e., `@main` returns anything other than `int`).
  * Probably make an issue about extending other implementations (other interpreters and compilers) with the new feature.

PR #305, opened 2023-12-11
AsyncBril

evanmwilliams:
Support for asynchronous programming in Bril


Pat-Lafon:
Hey! This is a really cool project! I'm glad that overall, the rust code base was fairly extensible for supporting your extension. I created a pr, #306, to help address one of your failing ci checks. I think the other is an easy `cargo fmt` to fix.

If you don't mind, I have a few suggestions/questions about your extension?

It would be helpful if your extension was testable with `make test/benchmark`(I think some of these files are really more tests than benchmarks). That should be as simple as adding some output files and adding the corresponding paths to the Makefile lists.

brilirs/src/check.rs
```diff
@@ -286,6 +294,7 @@ fn type_check_instruction<'a>(
           check_asmt_type(ty, &expected_arg.arg_type)
         })?;
 
+      // print to console
```
Pat-Lafon, line 297: Unclear?

brilirs/src/check.rs
```diff
@@ -362,6 +371,94 @@ fn type_check_instruction<'a>(
...
+      args,
+      funcs,
+      labels,
+      pos: _,
+    } => {
+      check_num_args(1, args)?;
+      check_num_funcs(0, funcs)?;
+      check_num_labels(0, labels)?;
+      check_asmt_type(&Type::Int, get_type(env, 0, args)?)?;
+      update_env(env, dest, op_type)
```
Pat-Lafon, line 387: I think you forgot to check the op_type here as is done in all the other cases?

brilirs/src/check.rs
```diff
@@ -400,7 +497,11 @@ fn type_check_instruction<'a>(
         Some(t) => {
           check_num_args(1, args)?;
           let ty0 = get_type(env, 0, args)?;
-          check_asmt_type(t, ty0)
+          if let Some(promise_t) = promise {
```
Pat-Lafon, line 500: It's unclear to me why the `promise` variable needs to be threaded through as an argument to `type_check_instruction`. Isn't this information available inside of `t` here? If so, that would be simpler

brilirs/src/interp.rs
```diff
@@ -2,8 +2,11 @@ use crate::basic_block::{BBFunction, BBProgram, BasicBlock};
 use crate::error::{InterpError, PositionalInterpError};
 use bril2json::escape_control_chars;
 use bril_rs::Instruction;
-
 use fxhash::FxHashMap;
+use std::collections::HashMap;
```
Pat-Lafon, line 6: FxHashMap should just be a drop in replacement no?

brilirs/src/main.rs
```diff
@@ -6,6 +6,7 @@ use std::fs::File;
 use std::io::Read;
 
 fn main() {
+  std::env::set_var("RUST_BACKTRACE", "1");
```
Pat-Lafon, line 9: FYI, this can be set in the command line without needing to modify the code.

docs/lang/async.bril
```diff
@@ -0,0 +1,81 @@
+Asynchronous Bril
```
Pat-Lafon, line 1: Is this extra file intended?

brilirs/src/interp.rs
```diff
@@ -201,12 +250,17 @@ impl fmt::Display for Value {
...
     }
   }
 }
 
-fn optimized_val_output<T: std::io::Write>(out: &mut T, val: &Value) -> Result<(), std::io::Error> {
+fn optimized_val_output<T: std::io::Write>(
+  out: &Mutex<T>,
+  val: &Value,
+) -> Result<(), std::io::Error> {
+  let mut out = out.lock().unwrap();
```
Pat-Lafon, line 263: Maybe a nit, but there isn't really a reason to change this function right? Can we acquire the lock first and then call this function as a separation of concerns? That is equivalent to what you need to do if you are using Display instead.

Pat-Lafon:
> Unresolved promises are automatically cleaned up when they go out of scope.

This is a fine design decision but at the risk of overstepping, I think this is unnecessarily restrictive. Right now, promises need to get cleaned up at the end of each function call, but intuitively, there isn't any reason they can't span function boundaries right? Another approach is for these promises to have similar semantics as memory, everything needs to be cleaned up by the time the program terminates.

brilirs/src/interp.rs
```diff
@@ -87,17 +91,61 @@ impl Environment {
   }
 }
 
+//A second heap for atomics, since AtomicI64 cannot be added to the Value enum without a lot of work
```
Pat-Lafon, line 94: Purely out of curiosity, did you consider making them a part of the Value enum more seriously? Let's say that Value didn't implement Copy and you modeled these atomics as a variant of `Arc<AtomicI64>` instead of this map, would that be equivalent? What would the implications of that design decision be?

brilirs/src/interp.rs
```diff
@@ -516,19 +622,58 @@ fn execute_value_op<T: std::io::Write>(
...
+      let arg0 = get_arg::<usize>(&state.env, 0, args);
+      //expected
+      let arg1 = get_arg::<i64>(&state.env, 1, args);
+      //update
+      let arg2 = get_arg::<i64>(&state.env, 2, args);
+      let res = state.atomics.compare_and_swap(arg0, arg1, arg2);
+      state.env.set(dest, Value::Int(res));
+    }
+    LoadAtomic => {
+      let arg0 = get_arg::<usize>(&state.env, 0, args) as usize;
```
Pat-Lafon, line 644: Unnecessary `as usize` cast. I think clippy has some suggestions that could be applied.

brilirs/src/interp.rs
```diff
@@ -172,10 +221,10 @@ enum Value {
   Float(f64),
   Char(char),
   Pointer(Pointer),
+  Atomic(usize),
```
Pat-Lafon, line 224: You might consider using the "NewType" idiom here to not mix this usize with the usize used for variable names.

brilirs/src/interp.rs
```diff
@@ -516,19 +622,58 @@ fn execute_value_op<T: std::io::Write>(
...
+  state: &mut State,
   op: bril_rs::EffectOps,
   args: &[usize],
   funcs: &[usize],
   curr_block: &BasicBlock,
   // There are two output variables where values are stored to effect the loop execution.
   next_block_idx: &mut Option<usize>,
   result: &mut Option<Value>,
+  out: Arc<Mutex<T>>,
+  stop_flag: Option<Arc<AtomicBool>>,
```
Pat-Lafon, line 676: This stop_flag is somewhat mysterious to me. Why do we need to do this when the program ends up panic-ing anyways?

I think this could be implemented as `&Option<AtomicBool>`.

Pat-Lafon: Could this fit into State?

Pat-Lafon:
> The heap is thread-unsafe, allowing multiple threads to concurrently read/write
to the heap without checking. This is done using Rust unsafe blocks.
Alternatively, we could've used a Rw<T> to lock/unlock the heap when threads
were modifying it, but we ultimately decided that it would be more performant
to not implement it this way and provide support for synchronization primitives,
allowing for more flexibility.

Hmmm, this seems scary. I'm concerned that it's not just thread-safety but that there is is also a possibility of memory corruption. Did you try profiling this to see what the penalty is? One observation, pointers can't be used to write to different memory blocks so you actually might be able to write to multiple memory regions simultaneously without contention. I guess you only need to take a whole heap write lock during alloc/free? Maybe something smart can be done here... memory allocation is by far brilirs's least optimized part.

brilirs/src/interp.rs
```diff
@@ -516,19 +622,58 @@ fn execute_value_op<T: std::io::Write>(
       let res = Value::Pointer(arg0.add(arg1));
       state.env.set(dest, res);
     }
+    Resolve => {
+      let res = match handlers.remove(&args[0]) {
```
Pat-Lafon, line 626: When I first saw this line, I thought it was a bug since you are the variable name(usize'd index) to find the handler which also happens to be a usize. I realize looking at the rest of this code that it probably isn't.

Still, this seems like a serious hit to modularity, each promise is tied to the variable name it is bound to so you can't call `id` on a promise, pass it as argument, or store it in an array. I think this should be documented. For what it's worth, I think this is a straight forward limitation to lift, though related to your scoping restriction somewhat.

brilirs/src/interp.rs
```diff
@@ -516,19 +622,58 @@ fn execute_value_op<T: std::io::Write>(
...
+        Some(handle) => handle.0.join().unwrap(),
+        None => panic!("Unexpected execution error"),
```
Pat-Lafon, line 628: Can we use an result error instead? With a more descriptive error message.

brilirs/src/interp.rs
```diff
@@ -481,13 +559,35 @@ fn execute_value_op<T: std::io::Write>(
...
+          ),
+        );
+      } else {
+        make_func_args(callee_func, args, &mut state.env);
+
+        let result = execute(state, funcs[0].clone(), out, stop_flag)?;
+
+        let val = match result {
+          Some(val) => val,
+          None => Value::Uninitialized,
```
Pat-Lafon, line 585: This result was originally an unwrap which I think better aligns with the program semantics. Typechecking should ensure that there is always a value here so it is not possible to create an Uninitialized value here

Pat-Lafon:
Sorry for all the comments, hopefully some of them were helpful but otherwise take them all with a grain of salt. Your team thought about this a lot more than I have and it's your project. 🦀

sampsyo:
Very cool implementation, everybody! @Pat-Lafon has some good advice here, if you're interested in continuing the hacking.

PR #306, opened 2023-12-11
bril2json clippy fix && Note LLVM version

Pat-Lafon:


sampsyo (approved):
Excellent as always; thanks!!

PR #308, opened 2024-01-04
Clippy 1.75 lint

Pat-Lafon:


sampsyo:
Good point, Clippy! :smiley:

PR #311, opened 2024-03-04
Add Leibniz formula for Pi benchmark

leewei05:
It took about 2 seconds to calculate the current approximation of Pi. If we add another `0` to the number of loops, it would take about 20 seconds to add another digit, so I just stop here.

1000000:
```
$ time bril2json < leibniz.bril | brili
3.14159365359077425
________________________________________________________
Executed in    2.28 secs    fish           external
   usr time    2.15 secs    0.52 millis    2.15 secs
   sys time    0.09 secs    1.62 millis    0.09 secs
```

10000000:
```
$ time bril2json < leibniz.bril | brili
3.14159275358980139
________________________________________________________
Executed in   20.98 secs    fish           external
   usr time   20.43 secs    0.44 millis   20.43 secs
   sys time    0.27 secs    1.43 millis    0.27 secs
```


sampsyo:
This is SUPER cool! Thank you!!

PR #312, opened 2024-03-15
Library-izing `brilift`

Alex-Fischman:
This PR:
- Moves code from `main.rs` to new file `translator.rs`
- Adds `lib.rs`
- Adds full `bril-rs` feature set to both `brilirs` and `brilift` to allow users to depend on both
- Resolves issues with adding those features in `translator.rs`

Note: I don't think it makes sense for the compiler to support the `speculate` and `ssa` features, so I marked them as `unimplemented!()`. In contrast, I think that it does make sense to support `char`, but I'm not confident in my ability to write C without memory leaks, so I marked that as `todo!()`.


brilirs/Cargo.toml
```diff
@@ -31,7 +31,7 @@ itoa = "1.0"
 [dependencies.bril-rs]
 version = "0.1.0"
 path = "../bril-rs"
-features = ["ssa", "memory", "float", "speculate", "char"]
+features = ["memory", "float", "ssa", "speculate", "position", "import", "char"]
```
Pat-Lafon, line 34: I was surprised that these features weren't added since `brilirs` supports most of these. I realized that they probably get pulled in during feature resolution with `bril2json` which does list these features.

sampsyo: That'd do it, I guess!

brilift/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+mod rt;
+pub mod translator;
+
+use crate::translator::{find_func, Translator};
+use bril_rs::Program;
+use cranelift_object::ObjectModule;
+
+pub fn compile(program: &Program, target: Option<String>, opt_level: &str, output: &str) {
```
sampsyo, line 8: It looks like `main.rs` now largely duplicates this function (for the `!args.jit` case). What do you think about turning that into a function call to this new `compile` function?

brilift/src/translator.rs
```diff
@@ -0,0 +1,1104 @@
...
+        let func_ptr = self.get_func_ptr(func_id);
+        let arg_ptrs = Self::val_ptrs(args);
+        let func = mem::transmute::<_, fn(*const *const u8) -> ()>(func_ptr);
+        func(arg_ptrs.as_ptr());
+    }
+}
+
+#[derive(FromArgs)]
+#[argh(description = "Bril compiler")]
+pub struct Args {
```
sampsyo, line 1066: Since the CLI args don't seem to be used here (only in `main`), perhaps this struct should move there?

sampsyo (commented):
Overall looks good! And the `unimplemented!` approach seems perfectly good for those features we're not covering.

I have a couple of very minor organizational suggestions—let me know what you think.

Alex-Fischman:
I've changed `Args` to `RunArgs` and added a `CompileArgs` struct to improve the API. I've also moved `RunArgs` into `main.rs` and made `main` call into `compile`.

sampsyo:
Awesome; looking good! It looks like Clippy has one somewhat weird but OK suggestion:
https://github.com/sampsyo/bril/actions/runs/8311293204/job/22745133930?pr=312#step:7:28

Something also seems to be up with the tests… simple stuff like `br.bril` seems to be hitting the `todo!()`s?
https://github.com/sampsyo/bril/actions/runs/8311293204/job/22745134011?pr=312#step:11:195

I can try to take a closer look, since I don't see an obvious reason why this would be triggering.

brilift/src/translator.rs
```diff
@@ -0,0 +1,1073 @@
...
+                returns: vec![],
+                call_conv,
+            },
+            Self::PrintFloat => ir::Signature {
+                params: vec![ir::AbiParam::new(ir::types::F64)],
+                returns: vec![],
+                call_conv,
+            },
+            Self::PrintChar => ir::Signature {
+                params: vec![ir::AbiParam::new(ir::types::R32)],
```
sampsyo, line 55: I think `R32` is a reference type… for characters, we probably want `I8`?

brilift/src/translator.rs
```diff
@@ -0,0 +1,1073 @@
...
+            },
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        match self {
+            Self::PrintInt => "_bril_print_int",
+            Self::PrintBool => "_bril_print_bool",
+            Self::PrintFloat => "_bril_print_float",
+            Self::PrintChar => todo!(),
```
sampsyo, line 90: I think this is the trouble with the tests! Probably the easiest thing here is just to implement the runtime function, which should be straightforward.

sampsyo (commented):
Looking into it a little closer, it seems like compilation crashes when trying to add the runtime functions! I've marked the place where this happens. It seems like maybe we should add a `_bril_print_char` function, which shouldn't be too complex? Or, we could just remove `PrintChar` from the relevant enum.

brilift/src/translator.rs
Alex-Fischman: Rust characters are 4 bytes wide and roughly correspond to C++ "wide characters". I had assumed that Bril's characters were the same way based on https://capra.cs.cornell.edu/bril/lang/char.html and the `bril-rs` `Literal` enum. I think that `I32` is the right thing in this case.

brilift/src/translator.rs
Alex-Fischman: I didn't do this because I don't know how to print Unicode characters from `printf`.

Alex-Fischman:
How do I run the CI?

sampsyo:
Unfortunately, the CI has to be manually triggered by a maintainer when the PR comes from a first-time contributor. It's a really annoying feature in GitHub that is sadly necessary to prevent widespread crypto scams:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Anyway, looking green from here! Thanks! I'll check things out more closely and then hit the button.

sampsyo:
OK, we're all set! Thanks again for getting this going!!!

I changed things around just a tiny bit: most prominently, making a `jit_run` interface in `lib.rs` that acts as a symmetric pair alongside `compile`.

PR #314, opened 2024-03-25
Expose the C runtime library to Rust library users

Alex-Fischman:
Adds a `c_runtime` function.

### DIFF
--- a/brilift/src/lib.rs
+++ b/brilift/src/lib.rs
@@ -71,3 +71,8 @@ pub fn jit_run(program: &Program, args: Vec<String>, dump_ir: bool) {
     // Invoke the main function.
     unsafe { trans.run(entry_id, &main_args) };
 }
+
+/// The C runtime library for Rust library users.
+pub fn c_runtime() -> &'static str {
+    include_str!("../rt.c")
+}
--- a/brilift/src/lib.rs
+++ b/brilift/src/lib.rs
@@ -74,5 +74,5 @@ pub fn jit_run(program: &Program, args: Vec<String>, dump_ir: bool) {
 
 /// The C runtime library for Rust library users.
 pub fn c_runtime() -> &'static str {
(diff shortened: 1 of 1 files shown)


sampsyo:
Ah, looks interesting! I can see potentially how this would be useful—any chance you could write a few sentences about the use case and why it's helpful to have this embedded in the library/executable instead of on the filesystem?

One somewhat annoying thing: the path `../rt.c` depends on the directory where `rustc` is run. It will run into problems when, for example, doing [cargo publish](https://github.com/rust-lang/cargo/issues/13309). I think the solution is to use the `$CARGO_MANIFEST_DIR` environment variable, which Cargo sets to the appropriate path. [This SO answer](https://stackoverflow.com/a/68865971) suggests one possible way, for instance. Can you give that a try and see if it works the way you expect?

Alex-Fischman:
This is useful because the alternative for library users is to go into this repo, download the file, and then move it into their repository. However, this will fail if this repo ever changes `rt.c`, because the local copy will not match. This is reasonably likely to happen in the future, for example if `char` becomes supported `rt.c` will need to add `_bril_print_char`.

Alex-Fischman:
It's also useful as a way to remind maintainers that `rt.c` is a necessary part of the API for `brilift`.

Pat-Lafon:
> This is useful because the alternative for library users is to go into this repo, download the file, and then move it into their repository. However, this will fail if this repo ever changes `rt.c`, because the local copy will not match. This is reasonably likely to happen in the future, for example if `char` becomes supported `rt.c` will need to add `_bril_print_char`.

Maybe not the best solution but if you include bril as a git-submodule and then add it via a path versus having cargo look at git directly, you can have a symlink to `rt.c` in your repository. It also exists at a deterministic location on your machine in `$HOME/.cargo/` though maybe that is a little less clean

PR #323, opened 2024-05-07
Clippy 1.78

Pat-Lafon:


sampsyo:
Excellent! Clippy is always so helpful, and so friendly. :smiley:

PR #324, opened 2024-05-10
Allow use of brillvm as library

oflatt:
This PR refactors rust brillvm a little to allow use as a library. It depends on pr #319, so the diff includes that work as well.


Pat-Lafon:
Hmm, this seems a little awkward to me. Are you doing this because you don't want to call brillvm via https://doc.rust-lang.org/std/process/struct.Command.html nor want to re-implement/specialize a version of the main file to your use?(The main part of `brillvm`, `create_module_from_program`, is publicly exposed in lib)

oflatt:
Yeah, we don't want to call brillvm via a process.
I see what you mean- we just need to call `create_module_from_program`. I'll do that instead.

Issue #328, opened 2024-07-08
 Suggestion to move `dot-product.bril` benchmark to `mem` directory


Lai-YT:
The benchmark program [dot-product.bril](https://github.com/sampsyo/bril/blob/main/benchmarks/core/dot-product.bril), which performs a dot product on two vectors using the memory extension of Bril, is currently placed in the `core` directory. Considering its reliance on the memory extension, it might be more appropriate to place this program under the `mem` directory instead of `core`.

> Thank you so much for the great CS 6120!!

sampsyo:
Thanks!! Seems right to me—any chance you wouldn't mind opening a small PR to move the benchmark?

Lai-YT:
Sure! I'll make a change on this. :ok_person:

PR #329, opened 2024-07-08
Move dot-product benchmark to mem directory

Lai-YT:
This benchmark uses the memory extension, thus it is better placed under `mem` directory.

> Please let me know if further modifications are required. :smile: 

Resolves: #328


sampsyo:
Excellent; many thanks!

PR #331, opened 2024-07-26
brilirs: error on duplicate labels

Pat-Lafon:
The following is kinda fun because `brili` terminates but `brilirs` does not(Whether one takes the first or last occurrence of a label).

```
@main{
    jmp .A;
    .A:ret;
    .A:jmp  .A;
}
```

While `brilirs` can mirror this behavior, it can also be "helpful" by returning an error when it detects multiple labels as the specification/users intention is ambiguous.

2 commits to clean up the function first


sampsyo:
Awesome; seems great. I'm not sure if this restriction is in the spec, but it probably should be. 🤪

Issue #342, opened 2024-10-22
[Brillvm] Handling returns in branches


oflatt:
Here's a Bril file that should work:
```
@abs(x: float): float {
  tmp431: float = const 0;
  tmp432: bool = flt x tmp431;
  br tmp432 .label433 .label434;
.label433:
  tmp436: float = const -1;
  tmp437: float = fmul x tmp436;
  ret tmp437;
  jmp .label435;
.label434:
  ret x;
  jmp .label435;
.label435:
}
````
Brillvm has logic that adds returns in the terminating block, and it always adds a return void. This causes this example to break because the return type is float.

I'm not sure what the fix is here exactly. Perhaps we need to detect returns and turn them into terminating blocks?

Pat-Lafon:
I think I have a cute solution by exploiting the fact that this is only possible in a dead block. I've created a pr for review.

Pat-Lafon:
Ah, I had to implement a fix for cranelift as well and maybe using a trap instruction is better...

PR #343, opened 2024-10-23
Fixes https://github.com/sampsyo/bril/issues/342

Pat-Lafon:
Unterminated, non-void basic_blocks weren't handled in brillvm. I think this is only ever an issue of dead code(for if this was reachable, then the Bril code would not be well-formed).

The simplest solution may actually be to insert a jump to some arbitrary block instead of trying to craft a well-typed return.


Pat-Lafon:
Seems like brilift also has this issue. The slight wrinkle is that jumps in cranelift take arguments. I've used a trap instead with an arbitrarily chosen trap code... which is maybe a more principled solution anyways.

bril-rs/brillvm/src/llvm.rs
```diff
@@ -1354,7 +1355,19 @@ pub fn create_module_from_program<'a>(
...
             // Make sure every function is terminated with a return if not already
             if !is_terminating_instr(&last_instr) {
-                builder.build_return(None).unwrap();
+                if return_type.is_none() {
+                    builder.build_return(None).unwrap();
+                } else {
+                    // This block did not have a terminating instruction
+                    // Returning void is ill-typed for this function
+                    // This code should be unreachable in well-formed Bril
+                    // Let's just arbitrarily jump to avoid needing to
```
oflatt, line 1364: 🤣

bril-rs/brillvm/src/llvm.rs
```diff
@@ -1294,28 +1294,29 @@ pub fn create_module_from_program<'a>(
                     }
                 });
 
-                (llvm_func, instrs, block, heap)
+                (llvm_func, instrs, block, heap, return_type)
```
oflatt, line 1297: could preserve original function here instead of instrs and return_type
That, or do a zip later with the original functions and the llvm ones

oflatt (approved):
You're the best, thanks!

sampsyo:
Awesome; good thinking here!! And trapping in the generated CLIF seems like a good idea. I'll poke around in the Cranelift docs to see if there's something else obvious to use here, just in case.

PR #349, opened 2025-01-26
Fixes minor typo in core.md

adityanathan:


sampsyo:
This is actually just colloquial silliness and not a mistake. :smiley:

PR #352, opened 2025-01-27
feat(bitcast): Add bit casting extension

ethanuppal:
> [!NOTE]
> The typescript compiler parses integers incorrectly, so I wasn't able to add the code for that (note to self for when the bug is fixed: you can see my attempts by looking through the early commits).

- Adds a documentation page for the bit casting extension.
- Adds a `bitcast` feature to `bril-rs`
- Implements bit casting in `brilirs`


sampsyo:
Seems at least plausible! The big question that this PR so far leaves unresolved is: which pairs of types are allowed? The current implementation in the Rust interpreter allows no-ops and int<->float. But I guess the idea is that this would be extensible? And different extensions can add their own allowed type pairs to the definition of `bitcast`?

A keep-it-simple solution here would be to just add `int2float` and `float2int` instructions and call it a day…

ethanuppal:
> The big question that this PR so far leaves unresolved is: which pairs of types are allowed?  

I guess that was what I meant to convey here:

> Takes one value of bit width $N$ and produces a value of the width-$N$ output
  type with the same bit representation, **if conversion is defined between the
  input and output types.** It is an error to bitcast between types of distinct 
  bit widths, or between types for which bitcasting is not well-defined.
>
> **Implementations/extensions may choose to define the semantics of bitcasting between their
custom types or between a custom type and a core type.**

---

> A keep-it-simple solution here would be to just add `int2float` and `float2int` instructions and call it a day…

Sounds good! I'll do that (although I'm not sure about that naming, since it seems to imply some sort of numeric approximation in either direction instead of a direct bitcast). I'm not sure if there's a formal way of specifying "dependencies on other extensions" (I can do this easily in Rust, though, by having `bitcast = ["float"]`.

sampsyo:
You're right; it seems like the names `int2float` and `float2int` would be better used for the lossy data-type conversions (which we should also probably add tbh?). Maybe `float2bits` and `bits2float`?

There is not really a way to specify dependencies between extensions… which is maybe why I am feeling drawn to a "worse is better" approach that wouldn't require figuring that general concept out.

ethanuppal:
@sampsyo Ready for review

brilirs/src/check.rs
```diff
@@ -261,6 +261,40 @@ fn type_check_instruction<'a>(
...
+      labels,
+      pos: _,
+      op_type,
+    } => {
+      check_num_args(1, args)?;
+      check_num_funcs(0, funcs)?;
+      check_num_labels(0, labels)?;
+      if !matches!((get_type(env, 0, args)?, op_type), (Type::Float, Type::Int)) {
+        return Err(InterpError::InvalidBitcastType);
+      }
```
Pat-Lafon, line 295: Nit: for consistency, I would prefer it if you didn't use `matches!` here and instead used the `check_asmt_type` function. A little cleaner at a glance and I like that it reports the erroneous type in the message.

brilirs/src/interp.rs
```diff
@@ -516,6 +518,32 @@ fn execute_value_op<T: std::io::Write>(
       let res = Value::Pointer(arg0.add(arg1));
       state.env.set(dest, res);
     }
+    Float2Bits => {
+      let input = get_arg::<&Value>(&state.env, 0, args);
+
+      let (Value::Float(float), Type::Int) = (input, op_type) else {
+        panic!("Invalid bitcast");
+      };
```
Pat-Lafon, line 526: Maybe this is a hold-over from your previous design but I don't think the explicit panic here is necessary anymore. The op_type is no longer needed at runtime after type-checking here. `get_arg` can get you the float directly

brilirs/src/interp.rs
```diff
@@ -516,6 +518,32 @@ fn execute_value_op<T: std::io::Write>(
...
+
+      // https://users.rust-lang.org/t/i64-u64-mapping-revisited/109315
+      // the to and from native endian stuff is a nop
+      // if link dies try web archive
+      let uint = float.to_bits();
+      let int = i64::from_ne_bytes(uint.to_ne_bytes());
```
Pat-Lafon, line 532: Nit: Just to check my understanding here, is it necessary to produce an intermediate u64 here as opposed to using `f64::to_ne_bytes`?

brilirs/src/check.rs
ethanuppal: Good idea!

ethanuppal:
Looks like this is failing on a new benchmark. Not sure why

brilirs/src/error.rs
```diff
@@ -59,6 +59,8 @@ pub enum InterpError {
   IoError(#[from] std::io::Error),
   #[error("value ${0} cannot be converted to char")]
   ToCharError(i64),
+  #[error("bitcasts must be between `int` and `float`")]
+  InvalidBitcastType,
```
Pat-Lafon, line 63: If this error case is no longer being raised then we can remove it.

Pat-Lafon:
> Looks like this is failing on a newly-merged benchmark. Not sure why since that benchmark doesn't touch `bits2float` or `float2bits`.

Yeah, this is unrelated. Seems like `brilirs` isn't aligned with `brili` on using scientific notation for certain floats large floats. I'll see about submitting a patch.

brilirs/src/error.rs
ethanuppal: Right, I was going to and forgot. Thanks for catching!

ethanuppal:
> > Looks like this is failing on a newly-merged benchmark. Not sure why since that benchmark doesn't touch `bits2float` or `float2bits`.
> 
> 
> 
> Yeah, this is unrelated. Seems like `brilirs` isn't aligned with `brili` on using scientific notation for certain large floats. I'll see about submitting a patch.

So I assume this is a blocker for this PR?

ethanuppal:
Now that #388 is closed, could CI please be rerun for this PR?

Pat-Lafon (approved):
My only suggestion from the implementation perspective would be to add your test case to the `make test` list of brilirs so that it's included in the github workflow

https://github.com/sampsyo/bril/blob/d41519bda841126cdd0c538e79f94921633c5565/brilirs/Makefile#L1-L7

ethanuppal:
Requesting review (cc @sampsyo)

sampsyo:
Looks good! Not sure when these arose, but there are some simple clippy lints that look worth addressing.

ethanuppal:
> Looks good! Not sure when these arose, but there are some simple clippy lints that look worth addressing.

No clue either, but I think it's happy now

ethanuppal:
I think your SSA changes have not fully been integrated into the Rust codebase, I guess we wait on an unrelated PR again :joy:

Pat-Lafon:
> > Looks good! Not sure when these arose, but there are some simple clippy lints that look worth addressing.
> 
> No clue either, but I think it's happy now

This happened to me as well once where the ci ran for the pr, a new rust version was released, and then the next ci run fails(in my case it was when my pr was merged to main).

I can see how it's annoying for other rust contributors(since this will only happen when the rust code changes). We can start tracking a specific Rust version in the ci instead of `stable` to avoid that issue... though it won't help you when there is a breaking change to a Bril extension.

Pat-Lafon:
> I think your SSA changes have not fully been integrated into the Rust codebase, I guess we wait on an unrelated PR again 😂

Yeah, yet again I have a patch in flight but it's not merged yet 🤷

ethanuppal:
> I can see how it's annoying for other rust contributors(since this will only happen when the rust code changes). We can start tracking a specific Rust version in the ci instead of `stable` to avoid that issue... though it won't help you when there is a breaking change to a Bril extension.

Fixing the Rust version is definitely a good idea 🙌

ethanuppal:
Maybe this PR could be set to automerge to reduce the probability that another change breaks the Rust code in between?

sampsyo:
Looks like we're good to go!

PR #356, opened 2025-01-29
BENCHMARKS: add sorting network benchmark

scober:
Add a benchmark that runs the optimal sorting network for a 5-input array. It was generated using the ts compiler so it is comically un-optimized.


scober:
Just added one! Sorry for the delay!

sampsyo:
Awesome; thanks!

PR #358, opened 2025-01-29
BRIL benchmark: General Binary Matrix Multiplication algorithm using packed matrix representation

UnsignedByte:
Adds a short algorithm to multiply two matrices represented by packed integers. Entries in the matrix are single bits.


sampsyo:
Looks like the type checker has a suggestion. Also, could you please add a brief description to the docs?

UnsignedByte:
Looks like I just had some issues with newlines, this should be ready now 🤞

sampsyo:
Excellent; thanks!

PR #359, opened 2025-01-29
[benchmark] add conway's game of life

gerardogtn:
This benchmark models Conway's Game of Life, provided two inputs n and seed and runs in two stages:
* First it generates a random nxn matrix, random numbers are generated using LCG with the seed input and prints it.
* Produces the next matrix according to the rules of Conway's Game of Life, and prints it.


sampsyo:
Wow; this is snazzy! Nice work. :smiley:

PR #362, opened 2025-01-30
[Benchmarks] Fast Inverse Square Root

mt-xing:
# Please note that this PR includes benchmarks that depend on an extension which provides `bits2float` and `float2bits`, which has not yet been merged.

This PR adds two benchmarks, called `fastinvsqrt` and `lamefastinvsqrt` respectively. Both will compute `1/sqrt(x)` using fast methods. The primary inspiration was the legendary [Quake 3 Fast Inverse Square Root Algorithm](https://en.wikipedia.org/wiki/Fast_inverse_square_root), which computes the value extremely fast, using some bit magic to get a logarithmic approximation and then uses a single iteration of Newton's method to refine the estimate. Unfortunately, this requires `bits2float` and `float2bits`, which a peer has submitted a PR to add to bril that has not yet been merged.

As an alternative, `lamefastinvsqrt` uses the much less interesting Newton's method, starting at `1/x` and iterating until the step size becomes smaller than a user-defined precision. This benchmark runs fine even on bril as it exists today. The TOML test case is thus provided for `lamefastinvsqrt` only.

It may be wise to hold off on merging this benchmark until the bit casting extension is merged.


sampsyo:
Sorry to be a stickler on a language thing here, but could you please rename "lame*" to something like "slow*" or "iterative*"?

mt-xing:
> Sorry to be a stickler on a language thing here, but could you please rename "lame*" to something like "slow*" or "iterative*"?

Sure! It's now called `iterativefastinvsqrt`

PR #363, opened 2025-01-30
combination benchmark

Annacaro22:
Compute binomial combination, ie. n choose k, for positive integers.


sampsyo:
Hello! It looks like the tests are failing because the interpreter couldn't parse the command-line arguments. I think you want `ARGS: 1 2` instead of `ARGS: 1, 2`?

Annacaro22:
My bad! Just fixed that.

sampsyo:
Thanks! Now the tests are failing because of the missing `.prof` output. That can also be generated with `turnt --save`; can you add that file?

Annacaro22:
Hmmm, when I run that command, I get the .out file, but no .prof is automatically generated. I have the -p flag in my turnt.toml flag. When I run in the command line with the -p flag, I get the total_dyn_inst printed to the terminal, so I'm not sure why it's having trouble making the file through turnt... my turnt.toml file is currently:
command = "bril2json < {filename} | brili {args} -p"
and i run it in the command line with: turnt --save combination.bril

Annacaro22:
I can copy paste it from the terminal into a .prof file? Would that work or no?

sampsyo:
You can certainly do it manually. Or you can just use `turnt --save` and the provided `turnt.toml`, which is configured to produce this file:
https://github.com/sampsyo/bril/blob/abb0d8170f41007988e92dbeb993a2c4cc90ceb4/benchmarks/turnt.toml#L4

Annacaro22:
That worked! Thank you so much.

sampsyo:
Looks good; thanks!

PR #364, opened 2025-01-30
add cordic benchmark

gabizon103:
this adds a benchmark that uses the cordic algorithm to compute the sine of 50 degrees


gabizon103:
yep, just added it!

sampsyo:
Great; thanks!

PR #365, opened 2025-01-30
Add sum-digits benchmark

bryantpark04:
This benchmark sums all of the base-10 digits of the input number.


sampsyo:
Looks good; thanks!

PR #366, opened 2025-01-30
[Benchmark] Connected Components in Undirected Graph 

KabirSamsi:
This presents a benchmark which introduces an abstraction for stacks in memory with `push` and `pop` operations, and uses these to implement an iterative DFS-based graph traversal to print out and count each of its connected components. More in my discussion writeup!


benchmarks/mem/bubblesort.prof
sampsyo, line 1: Looks like an unrelated change; can you revert this one?

sampsyo (commented):
Wow; very ambitious! Nice work!

It looks like the tests are failing because the program needs an `ARGS:` comment so Turnt knows which CLI arguments to provide. Could you please add that?

KabirSamsi:
Sorry about that! Both the bubblesort diff and the `ARGS` flag should be fixed now.

sampsyo:
Hmm; the tests are still failing (this time because of mismatch output). Maybe try `turnt --save`?

KabirSamsi:
I'll try that out.

KabirSamsi:
I don’t see a confirmation that the CI has passed, but all turnt tests are working on my end now! I realized I had to fix a trailing line in the profiling file.

sampsyo:
Looks good; thanks!!

PR #370, opened 2025-01-30
docs(editor): Add new editor plugins

ethanuppal:
Added the VSCode and Neovim wrapper plugins I made to the editor docs. I can remove my self-attribution if desired.

### DIFF
--- a/docs/tools/plugin.md
+++ b/docs/tools/plugin.md
@@ -17,7 +17,7 @@ If you're using Neovim, Ethan Uppal made a simple wrapper around `bril-vim` for
 the [Lazy](https://lazy.folke.io) package manager, supporting LSP in addition to
 the syntax highlighting (link [here](https://github.com/ethanuppal/bril.nvim/tree/main)):
 
-```
+```lua
 { "ethanuppal/bril.nvim" }
 ```
 
--- a/docs/tools/plugin.md
+++ b/docs/tools/plugin.md
@@ -1,9 +1,10 @@
 Syntax Plugin for Text Editors
 ==============================
 
-## VSCode
+## Visual Studio Code
 
-You can install the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=EthanUppal.bril), made by Ethan Uppal.
+You can install the [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=EthanUppal.bril), made by Ethan Uppal.
+Follow the instructions in the extension README.
 
 ## (Neo)Vim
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Seems good! What do you think about adding the Neovim plugin to this monorepo (so others can maintain it alongside other Bril stuff)? Would that be overkill?

ethanuppal:
I think it would be overkill since it just loads the vim plugin and starts the LSP, which you can easily do manually but its nicer to have a one liner.

docs/tools/plugin.md
```diff
@@ -1,12 +1,26 @@
 Syntax Plugin for Text Editors
 ==============================
 
+## VSCode
```
sampsyo, line 4: Let's spell it out: "Visual Studio Code."

docs/tools/plugin.md
```diff
@@ -1,12 +1,26 @@
...
+
 There is a [Vim][] syntax highlighting plugin for Bril's text format available in `bril-vim`. You can use it with a Vim plugin manager. For example, if you use [vim-plug][], you can add this to your `.vimrc`:
 
     Plug 'sampsyo/bril', { 'for': 'bril', 'rtp': 'bril-vim' }
 
 You can read [more about the plugin][blog], which is originally by Edwin Peguero.
 
+If you're using Neovim, Ethan Uppal made a simple wrapper around `bril-vim` for
+the [Lazy](https://lazy.folke.io) package manager, supporting LSP in addition to
+the syntax highlighting (link [here](https://github.com/ethanuppal/bril.nvim/tree/main)):
```
sampsyo, line 18:
```suggestion
If you're using Neovim, Ethan Uppal made a [simple wrapper around `bril-vim`][bril.nvim] for
the [Lazy](https://lazy.folke.io) package manager, supporting LSP in addition to
the syntax highlighting:

[bril.nvim]: https://github.com/ethanuppal/bril.nvim/tree/main
```

Use an actual in-context link.

docs/tools/plugin.md
```diff
@@ -1,12 +1,26 @@
...
+
+```lua
+{ "ethanuppal/bril.nvim" }
+```
```
sampsyo, line 22: This doesn't quite have enough context. Can you add, like, one half sentence telling the user what they are supposed to do with this code block?

sampsyo (changes requested):
I don't really understand the argument for why a separate repo is better for trivial things (seems to me like the opposite is true, i.e., big complicated projects maybe shouldn't get monorepo-ized). But whatever, it doesn't really matter! Here are some docs suggestions.

ethanuppal:
I guess the VSCode plugin could be easily added. However, since Lazy makes it hard to deal with runtime paths, there plugin files would have to be relative to the project root and not a directory, which is probably less than desirable. I'd be happy to monorepoize either, though.

sampsyo:
> the plugin files would have to be relative to the project root and not a directory

That's a good reason!

PR #372, opened 2025-01-30
added ray-bbox-intersection benchmark

mariasoroka:


sampsyo:
Looks great; thanks!!

PR #378, opened 2025-01-31
Add Blum Blum Shub benchmark

InnovativeInventor:
This benchmark prints the least significant bit of each elm of a given BBS sequence. To spot-check for correctness, the tested program output sequence is matched against example output in the [Wikipedia entry](https://en.wikipedia.org/wiki/Blum_Blum_Shub) for the BBS PRNG (when p=11, q=23, seed=3, and sequence_length=6). 

No additional testing of this program (e.g. on other BBS sequences) was performed.


sampsyo:
Looks good; thanks!

PR #380, opened 2025-01-31
Add McCarthy 91 function benchmark

mb64:
The McCarthy 91 function is a classic benchmark recursive function with interesting recursive behavior (it's nested recursion, like the Ackermann function) but simple output behavior -- it returns 91 for all small inputs.


sampsyo:
What a strange and interesting function! Nice find.

PR #383, opened 2025-01-31
Karatsuba multiplication benchmark

smd21:
Implements karatsuba's multiplication algorithm


sampsyo:
Hi, Serena! Incredibly, it looks like @arnavm30 had a similar idea, in #376. I think it's fine to have two implementations of this, but would you mind renaming your files to something else? I dunno, `karatsuba2`? But if you take a look at the two programs and decide they're pretty close to each other, it's fine to just keep one as well (with no grade implications, obvi).

PR #387, opened 2025-02-02
Add hamming benchmark

lisarli:
This benchmark computes the hamming distance between two integers.


sampsyo:
Looks good; thanks!

PR #391, opened 2025-02-05
add rot13 benchmark

aw578:
should be fixed now!


sampsyo:
Thanks! This might win the award for shortest benchmark…

PR #392, opened 2025-02-05
Move bencmarks in core to relevant directories

emmanueljs1:
This PR moves the `birthday.bril` to `float/` because it uses the `float` type, and `char-poly.bril` to `mem/` because it uses pointers.


sampsyo:
Thanks, @emmanueljs1!!

PR #393, opened 2025-02-05
[bril-ocaml] Count `ptr`s in `count.ml`

polybeandip:
Tweaks `count.ml` to correctly display number of `ptr`s instead of always printing `0`.


sampsyo:
Good catch!

PR #396, opened 2025-02-06
Allow for array of glob patterns in brench

UnsignedByte:
Allows for `benchmarks = [a, b, c]`, a list of different glob patterns to match files in brench.


brench/brench.py
```diff
@@ -87,8 +87,11 @@ def brench(config_path, files, jobs):
         config = tomlkit.loads(f.read())
 
     # Use configured file list, if none is specified via the CLI.
-    if not files and 'benchmarks' in config:
-        files = glob.glob(config['benchmarks'], recursive=True)
```
sampsyo, line 91: Looks like `recursive=True` got dropped in the `glob` call.

sampsyo (changes requested):
Looks good overall; I caught little detail that got lost in the change!

brench/brench.py
UnsignedByte: Done!

PR #398, opened 2025-02-07
Scientific floats

### LINKED ISSUE #388: [brili/brilirs] Clarifying printing large floats
The float extension specifies that:
`
The [core print operation](https://capra.cs.cornell.edu/bril/lang/core.html#miscellaneous) prints float values with 17 decimal digits of precision, including trailing zeros. (This is like using the %.17lf format specifier in C’s printf.)
`

Consider a minified example based on #373
```ocaml
@main {
  v0: float = const 1e21;

  print v0;
}
```

`brilirs` returns
```
1000000000000000000000.00000000000000000
```

modifying `brilirs` to have `e` in the format args can produce
```
1.00000000000000000e21
```

`brili` currently produces
```
1e+21
```
(The unexpected thing here may be that `toFixed` in Javascript falls back to the `toString` implementation on 1E21 and larger floats https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#description)

I would like to clarify what the intended output should be so I can submit a patch to address the failing test case observed in #352. The current `brilirs` behavior seems to me to be a little closer to the spec. I am not opposed to using exponential notation for printing floats but there might be some things to iron out to get matching output. This might of come up before in https://github.com/sampsyo/bril/pull/212 though I punted on it at the time. For example, should there be a `+` between the `e` and the number as in `brili`'s output or without as in the modified `brilirs` output?

Pat-Lafon:
Fixes #388

I've bundled in two additional fixes so far. The rust `bril2json` implementation treats `from` as a keyword and wasn't accepting it also as an identifier. I have also moved the random_walk benchmark to mixed so that brilift can ignore it as it uses both the mem and char extensions.

My attempt to address #388 has all floating points with abs(exponent) >= 10 use exponential notation. This is the minimum I would set it to, to avoid language differences in whether to require 2 digits after the e like c's printf does (like `1e09`).

Since Rust was the main outlier in not extending positive exponents with a `+`, it seems reasonable to also expect that for numbers printed as an exponential (like `1.1e+11`). 

Given the variation in how significant digits with an exponent are treated(and some odd floats when the number's exponent is smaller than the number of significant digits), it seems more uniform to not require a specific number of significant digits when using `e`(which ends up being equivalent to %g in c's `printf`).

I also slightly touched up `brili`'s handling of `-0.0` for which Javascript is the only language of those I tried that wanted to print positive zero.

I would like to put up this pr for discussion, there are different choices that can be made here. For example, maybe printing floats should just be `%g` compatible and require that implementations pull in a c-printf compatible formatter. The languages I've looked at are Javascript, Rust, C, and Ocaml.

I have not updated the documentation yet.


Pat-Lafon:
So correction, seems like `%g` will sometimes cap the number of significant digits at 6... so the number of significant digits in the exponent case should probably be fixed instead of allowed to be variable.

brili.ts
```diff
@@ -594,7 +594,8 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
     let args = instr.args || [];
     let values = args.map(function (i) {
       let val = get(state.env, i);
-      if (Object.is(-0, val)) { return "-0.00000000000000000" };
+      if (typeof val == "number" && Object.is(-0, val)) { return "-" + val.toFixed(17) };
+      if (typeof val == "number" && val != 0 && Math.abs(Math.log10( Math.abs(val))) >= 10) { return val.toExponential() }
```
sampsyo, line 598: Sorry if I'm being slow here, but I don't 100% grok the impact of this change—is it a more principled way to get the same behavior (without hard-coding the string for the negative zero case, to work around the quirk in `toFixed`), or has the behavior for negative zero changed?

It might be nice to refactor this stanza at this point to look like this:

```js
if (typeof val == "number") {
  if (Object.is(-0, val)) {
    ...
  } else if ... {
    ...
  } else {
    ...
  }
}
```

sampsyo (commented):
Awesome; thank you for getting started. Overall, I like your idea here: namely, pick a value threshold and switch between scientific notation and, uh, "non-scientific" notation. But in either case, keep the number of decimal points the same.

You seem to be right that C's `%g` specifier is similar in spirit (try to switch between `%f`-like and `%g`-like formatting) but not exactly the same. So instead of using `%g`, maybe we should just use `%e` directly for numbers that exceed the threshold?

brili.ts
Pat-Lafon: I don't think it's change any behavior since an integer can't be a -0(I hope). I've implemented your suggestion which is clearer.

brili.ts
```diff
@@ -594,8 +594,12 @@ function evalInstr(instr: bril.Instruction, state: State): Action {
     let args = instr.args || [];
     let values = args.map(function (i) {
       let val = get(state.env, i);
-      if (Object.is(-0, val)) { return "-0.00000000000000000" };
-      if (typeof val == "number") { return val.toFixed(17) } else {return val.toString()}}
+      if (typeof val == "number") {
+        if ( Object.is(-0.0, val)) { return "-" + val.toFixed(17) }
+        else if (val != 0.0 && Math.abs(Math.log10(Math.abs(val))) >= 10) { return val.toExponential(6) }
```
sampsyo, line 599: I guess this is the crux of where we need to make a decision: should this be `toExponential(17)` to match the precision of the non-scientific-notation case?

PR #399, opened 2025-02-08
feat: Add another `bril2json` implementation

ethanuppal:
### Content

This PR adds a `bril2json` implementation based on my frontend <https://crates.io/crates/bril-frontend>. It adds a new directory, `other/`, and under it a Rust binary crate called `bril2json`. Inside `other/bril2json/`, you will find:

- `test.py`: A simple parallel test runner I used to confirm it correctly parses all of the benchmarks
- `src/main.rs`: CLI driver that takes in either one filename argument or standard input
- `src/lib.rs`: Library to convert `bril-frontend`'s AST into JSON

One advantage of this tool over existing tools is that it has extra nice diagnostics (made possible because `bril-frontend` keeps track of lots of location information):

image

> [!NOTE]
> Looks like there's a typo in `bril-frontend@0.2.0` where the error message for effect operations refers to value operations. I've fixed that in `bril-frontend@0.2.1` and updated `other/bril2json/Cargo.toml` correspondingly.

### Other

I know the PR doesn't have these things right now, but I could work on the later:

- `bril-frontend` wraps all AST structures in a special `Loc` deref type, which can be used for all the position information in the position extension.
- `bril-frontend` could be extended to support more extensions beyond core, float, char, and mem, which would mean that this `bril2json` could be extended likewise.
- To keep this PR small, I've omitted documentation changes.


ethanuppal:
A blocking TODO is to relicense `bril-frontend` under an MIT-compatible license (I had made it GPL without much thinking, so this is currently GPL too, which doesn't work in an MIT codebase).

sampsyo:
Cool; this seems like a fun addition!

Here are some organization thoughts, at the risk of delaying the merging of this further (and maybe that is a reason to work on these ideas after this PR is merged). It seems kinda odd for this tool to be in the monorepo, but for its main "workhorse" crate not to be. Among other reasons, it would prevent the versioning shenanigans you already mentioned above. What do you think about first monorepo-izing the `bril-frontend` dependency?

One way to make that more palatable might be to say that, for the semester, you might want to develop most of that dependency in your own fork of the Bril repo, and then afterward, things will presumably have settled down a bit and it will be clearer what we want to merge.

Another would be to consider breaking your `bril-frontend` dependency into smaller crates. With this PR as an example, it looks like the standalone `bril2json` tool doesn't actually need the type inference step? Or maybe it does? Anyway, it could be an interesting exercise to think about decomposing as far as possible to allow mix-and-match dependencies.

Let me know what you think!

sampsyo:
Oh, and also, on the extremely minor question of the directory: I don't think we should create an "other" directory. (Other than what?) If this does land as a standalone thing, let's just put it in the top level under the name `bril2json-rs` or similar.

ethanuppal:
> One way to make that more palatable might be to say that, for the semester, you might want to develop most of that dependency in your own fork of the Bril repo, and then afterward, things will presumably have settled down a bit and it will be clearer what we want to merge.

Yes, that's what I was thinking:
1. To get `bril-frontend` merged into main Bril would likely require a long PR or multiple PRs splitting it up, which would take a while.
2. Once it's merged, I wouldn't be able to make changes (like correcting the error message) in seconds --- it would take days and a PR.
Once it's in a stable state, I'll archive my repo and contribute the changes here.

ethanuppal:
> Oh, and also, on the extremely minor question of the directory: I don't think we should create an "other" directory. (Other than what?) If this does land as a standalone thing, let's just put it in the top level under the name `bril2json-rs` or similar.

I meant to imply this is "yet an**other**" implementation, separate from the canonical Python or existing Rust implementation.

ethanuppal:
Requesting review (@sampsyo)

sampsyo:
Can you please remove the `other` directory? I would like to go with my proposal above, i.e., a top-level `bril2json-rs` directory for the Cargo project. I don't think it's a good idea to start a trend toward partial categorization like this.

ethanuppal:
> Can you please remove the `other` directory? I would like to go with my proposal above, i.e., a top-level `bril2json-rs` directory for the Cargo project. I don't think it's a good idea to start a trend toward partial categorization like this.

It is done

sampsyo:
Looks good; thank you!!

Another thing to think about for a potential follow-up: configuring Turnt for differential testing against the reference parser.

Issue #330, opened 2024-07-23
`phi` instructions have the "swap problem"


sampsyo:
@pervognsen [pointed out on Mastodon](https://mastodon.social/@pervognsen/112794867961757636) that `phi` instructions behave in a weird and probably bad way. Because they are "just normal instructions," they take effect immediately, i.e., all subsequent instructions immediately see the update to the variable they write. That includes other `phi` instructions that use the same variable on their right-hand sides.

Here's a self-contained example showing this behavior:

```
@main {
  i: int = const 5;
  one: int = const 1;
  zero: int = const 0;

.l0:
  x0: int = const 0;
  y0: int = const 1;
  jmp .l1;

.l1:
  x1: int = phi .l0 x0 .l1 y1;
  y1: int = phi .l0 y0 .l1 x1;
  print x1 y1;

  cond: bool = gt i zero;
  i: int = sub i one;
  br cond .l1 .end;

.end:
}
```

Under Bril's current semantics (as implemented by the reference interpreter and my `from_ssa.py` example), the two `phi`s execute in order, so the first one clobbers `x1` before the second one reads it.

This behavior is nonstandard and problematic. Under a more normal SSA semantics, the `phi`s would execute "simultaneously," i.e., they would all read the RHS values from the start of the basic block, never from earlier `phi`s in the same basic block. This program would then swap the values of `x1` and `y1` on every trip around the loop.

In the context of out-of-SSA transformations, this issue is called the "swap problem" in [this classic Preston Briggs banger](https://homes.luddy.indiana.edu/achauhan/Teaching/B629/2006-Fall/CourseMaterial/1998-spe-briggs-ssa_improv.pdf).

Anyway, my take here is that this (along with #108) is yet another consequence of my trying too hard to treat φ-nodes as "just normal Bril instructions." I was enamored with the idea that Bril's SSA form could be a tiny extension on top of the baseline non-SSA language, that interpreters wouldn't have to work too hard to tack on this one additional feature, and that we wouldn't need awkward restrictions like having all the φ-nodes appear at the beginnings of basic blocks. I think this has worked out poorly, and an SSA variant needs deeper changes to the language. At the very least, we cannot treat φ-nodes as normal instructions that read their arguments and write their results in order, like any other instruction.

[...]

PR #403, opened 2025-02-20
Design a new SSA extension

sampsyo:
I've tried to redesign the Bril SSA extension to deal with some long-standing issues, such as #330 and #108.

I had already been thinking about trying basic block arguments, but with a twist where the "sending side" of a phi-node was a separate instruction. And then I saw [Filip Pizlo's SSA variant](https://gist.github.com/pizlonator/79b0aa601912ff1a0eb1cb9253f5e98d), which is essentially a full realization of that fuzzy idea. So this new version of the extension works that way: there are both `phi` and `upsilon` instructions, and the semantics is now decoupled from the control flow. Aside from the benefits that Fil outlines in his note, this is generally a lot nicer for retrofitting SSA without disrupting the original Bril language too much; for instance:

* No problems with "simultaneous effects" of phi-nodes, nor restrictions on where phi-nodes may appear. So `phi` and `upsilon` behave quite a bit more like "normal" Bril instructions.
* The Bril `phi` instruction was really weird to implement in an interpreter because it involved tracing the most recently-executed labels.
* Converting out of SSA form is now even more trivial to implement in a naive way.

I also bit the bullet and added an `undef` instruction and a special *undef* value, which finally addresses #108. The summary is that it the only thing you are allowed to do with the undef value is to copy it around with `id` (or `phi` or `upsilon`); you can't (for example) feed it into an `add`. Some consequences include:

* Replacing `foo: t = undef;` with `foo: t = const x;` for any `x` is a valid refinement (and a sensible strategy for compiling the instruction).
* All `undef` instructions must be transitively dead. The jury is still out about how easy it is to write an analysis that determines this deadness.

The reference interpreter supports the new SSA instructions, including `undef`. I have dropped support for the old `phi` instruction. I have not updated any of the other interpreters/compilers.

Finally, I have implemented a naive to-SSA conversion that follows the "really crude approach" from [Appel's "SSA is Functional Programming](https://www.cs.princeton.edu/~appel/papers/ssafun.pdf). In that naive version, you create a version of every single variable for every single basic block, whether the block uses the variable or not. That's a lot of variables, but it makes for a very clear correctness argument.

I have not yet updated my dominance-frontier-based SSA conversion; I'll try that next once I'm confident this actually works.


sampsyo:
I think I'm happy with this as a baseline for future improvements! I am still thinking about changing the opcodes, per [this thread](https://discuss.systems/@jamey@toot.cat/114039120527133965), but that (among many other things) is best done in a separate PR.

PR #404, opened 2025-02-21
Rust ssa2

Pat-Lafon:
Updates the `bril-rs` and `brilirs` support of the ssa extension given the new design in #403. Drops support for the SSA extension in `brillvm`.

I will probably need to rebase again when #403 merges to slightly clean up the history.


Pat-Lafon:
This pr is well-timed with the release of the Rust 2024 edition. I have added a commit for this update.
- Mostly `cargo fmt` changes which are now allowed to occur over an edition boundary
- More unsafe things are explicitly labeled as unsafe, for this change it was:
    + https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html
    + https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html

sampsyo:
INCREDIBLE speed on this, @Pat-Lafon! :sparkles: I am very impressed.

I hit the "rebase" button right on GitHub; let's see if that worked…

bril-rs/brild/src/lib.rs
```diff
@@ -55,7 +55,7 @@ fn mangle_instr(code: AbstractCode, name_resolution_map: &HashMap<String, String
                 .map(|f| {
                     name_resolution_map
                         .get(&f)
-                        .expect("Could not find name for {f}")
+                        .unwrap_or_else(|| panic!("Could not find name for {f}"))
```
sampsyo, line 58: Seems like a somewhat odd change—it seems like the same thing but a little longer? Is this a new thing that Clippy is encouraging in some contexts, for whatever reason?

sampsyo (approved):
Looks great!! I took a look at everything and am very much on board. I had one extremely inconsequential comment on a style change. Happy to merge whenever you're ready.

bril-rs/brild/src/lib.rs
Pat-Lafon: This was actually two lints.

The newer one catches that `.expect("Could not find name for {f}")` is trying to be a format string but there is actually no macro to process the format string args. The error message is literally `Could not find name for {f}`. 

The fix is to use `.expect(&format!("Could not find name for {f}"))` except that now eagerly does some allocation and string manipulation for an unlikely error path, so it's more efficient(though this probably doesn't matter for `brild`) to thunk-ify the string formating.

sampsyo: Pretty weird! But sounds like a fine workaround.

Issue #341, opened 2024-10-22
[rs2bril] Duplicate variable names with different types compiled incorrectly.


oflatt:
Rust files that re-use a name with a different type cause the rs2bril compiler to generate ill-typed bril.

Example:
```
fn main() {
  let x: f64 = 2.0;
  let x: i64 = 2;
}

```

Pat-Lafon:
What bril code would you expect this to generate?
```bril
@main() {
  x: float = const 2.0;
  x: int = const 2;
}
```

We could generate this, but:
```sh
❯ bril2json < dupe.bril | brilck
new type int for x conflicts with old type float
```

(This might be a limitation of the type checking of brilck/brilirs to be fair)

oflatt:
I wasn't sure if it's a limitation of type checking or an intended restriction.
One way to get around this is to generate fresh names for each x

Pat-Lafon:
Hmm, it could work, but note that fresh names are tricky because I will now need to construct a control flow graph to understand what to re-write later uses of x to. Maybe one can show in most/all well-typed rust programs that one x dominates the other?

Somewhat unrelated but a fun example of bril typing that is probably not well supported(except by brili)
```
@main(cond: bool) {
  br cond .there .here;
.here:
  x: int = const 2;
  jmp .end;
.there:
  x: bool = const false;
.end:
  print x;
}
```

oflatt:
Could we do the renaming over the rust ast?

Alternatively a lightweight solution for now that would be a lot better is to throw an error

Pat-Lafon:
Hmm, I've punted and taken your solution of just throwing an error. Hopefully this isn't too limiting in terms of what rust programs are not accepted. There are probably better solutions.

oflatt:
I'm perfectly happy with that solution, thanks for putting so much effort into improving it!

PR #406, opened 2025-02-23
Dynamic bril

Pat-Lafon:
Since I'm in the swing of things, I would like to propose a (deceptively?) simple extension of adding the `any` type to Bril.

This could provide a more satisfactory solution to #341.

What I find cool about this extension is the potential for (much less efficient) implementations of higher level abstractions like #107 #138 and https://www.cs.cornell.edu/courses/cs6120/2020fa/blog/hof/ with minimal tooling changes.

See `docs/lang/dynamic.md`(and my PoC tests) for more details.


Pat-Lafon:
(Waiting on #404 to fix ci)

docs/lang/dynamic.md
```diff
@@ -0,0 +1,83 @@
...
+
+```
+v: any = const 4;
+b: bool = and v v; # Attempting to use an int as a bool
+```
+
+The `main` function cannot have any arguments with the `any` type.
+
+Interactions with float extension
+----------------------------------
```
sampsyo, line 40: Here's how I would describe what is going on here:

* For `const` instructions, the type of the instruction typically defines how to interpret the `value` field.
* When the destination type of a `const` instruction is `any`, we obviously can't do that. One option would have been to just disallow `x: any = const ...`. Instead, we opt to allow it using a limited set of types that can be uniquely identified by the JSON type of the `value` field.
* This strategy for `x: any = const ...` only works for Bril types that can be unambiguously associated with JSON value types. We associate JSON floating-point literals with the `float` Bril type and integer literals with the `int` type.
* This can lead to one confusing corner case where an `x: float = const ...` instruction, when "erased" to be `x: any = const ...`, takes on a different (dynamic) type.

Or we could just sidestep this by either changing the way that `float` constants work (i.e., require that they be serialized with a `.0` in the JSON representation), or disallowing `x: any = const ...` and requiring type annotations for constant instructions.

docs/lang/dynamic.md
```diff
@@ -0,0 +1,83 @@
...
+i: int = const 4;
+a: any = const 4;
+f: float = const 4;
+print i a f; # `4 4 4.00000000000000000`
+```
+
+Interactions with memory extension
+----------------------------------
+
+The memory extension expects `a uniformly-typed region of values` which is
```
sampsyo, line 63:
```suggestion
The memory extension expects "a uniformly-typed region of values" which is
```

sampsyo (commented):
This is a really nifty idea; I like it! Among other things, it does indeed serve as a convenient escape hatch for prototyping semantics that are not well suited to Bril's very-basic type system.

I mainly have some comments on how to document the purpose and quirks of the type system.

docs/lang/dynamic.md
Pat-Lafon: Hmmm, this can be changed to one of those other options if you have a preference. It just so happened that this behavior slotted perfectly into the current implementation.

I'll point out it's maybe not as strange as it first appears. Rust has somewhat similar behaviors where it will default an integer to be `i32` if it is otherwise unconstrained https://rust-lang.github.io/rfcs/0212-restore-int-fallback.html

sampsyo: I think it's fine to stick with this as you've implemented it! We can always go back and change it later (the easiest thing to fall back to, of course, would be just disallowing `any`-typed constants altogether, which think wouldn't be a great shame).

sampsyo:
This all looks good!! Would you say this is ready to merge?

Pat-Lafon:
> This all looks good!! Would you say this is ready to merge?

The only thing I haven't addressed is your comment about the "ideal use case" in the documentation.

I of course agree that using this extension to implement basically any of the features I list would be both horribly inefficient and (somewhat purposefully) provides no guard rails to help someone use it correctly(as I didn't implement a language-level way to check what the runtime type of an `any`-typed value is).

I could put that as a disclaimer in the documentation. I don't have a great sense for what kinds of dynamic language features would make this a good target.

sampsyo:
Cool cool, sounds good! I'll give the documentation one editing pass in that case and then merge.

About compiling dynamic language features: I guess what I was thinking is that, if you wanted to compile (e.g.) Python or JavaScript to Bril, then compiling to all `any` types would be a natural fit (because you don't know the types of any variables anyway). Could be a fun project.

sampsyo:
It also occurs to me to mention: one thing we could expand here (after this PR) is a specification for what is supposed to happen at run time. For most errors in Bril, doing something wrong (such as referring to an undefined variable) is undefined; implementations are supposed to do whatever they want to assume this doesn't happen. For this extension, we probably want to define the behavior to require exiting with an error, i.e., to remain sound by requiring implementations to perform dynamic type checking.

sampsyo:
Docs tweaked and expanded; I'll hit the merge button now!

Pat-Lafon:
> Cool cool, sounds good! I'll give the documentation one editing pass in that case and then merge.
> 
> About compiling dynamic language features: I guess what I was thinking is that, if you wanted to compile (e.g.) Python or JavaScript to Bril, then compiling to all `any` types would be a natural fit (because you don't know the types of any variables anyway). Could be a fun project.

The documentation looks very much improved to me! Many thanks

Pat-Lafon:
> It also occurs to me to mention: one thing we could expand here (after this PR) is a specification for what is supposed to happen at run time. For most errors in Bril, doing something wrong (such as referring to an undefined variable) is undefined; implementations are supposed to do whatever they want to assume this doesn't happen. For this extension, we probably want to define the behavior to require exiting with an error, i.e., to remain sound by requiring implementations to perform dynamic type checking.

Ah that is interesting. That is usually the bargain between "well-formed" Bril programs and tooling. I can certainly see adding it, though I wouldn't want to too seriously slow down brili/brilirs/etc for a maybe obscure extension. Part of this might be finding a light-weight way to opt-in/out of dynamic type checking.

sampsyo:
Yeah, that's a good point too. Both styles of semantics could be useful in different contexts.

PR #407, opened 2025-02-24
Rename "new SSA" operations to `set` and `get`

sampsyo:
On Mastodon, @jameysharp convinced me that it would be a good idea to rename the ops in "Pizlo SSA" form from `upsilon`/`phi` to something more meaningful. It's not like `phi` was a memorable name to begin with, and adding a second Greek letter makes it even more obscure; a break with tradition is a good time as any to make up names that would be easier to explain and remember.

I went with `set` and `get`.

So far, I've started with a mass find-and-replace; I'm sure this was too aggressive in parts and I'll need to roll back some changes that are premature. The funniest one, which I already avoided before the first commit here, was an attempt to change the copyright line in `hashmap.c` to attribute Jean-Getlippe Aumasson instead of Jean-Philippe Aumasson.


sampsyo:
OK, I think I've cleaned everything up and we're now in a consistent `set`/`get` state!

PR #409, opened 2025-02-25
Update "proper" SSA conversion for ssa2

sampsyo:
In #403, I added a "crude, bad" SSA conversion for the new ssa2 extension, but I kicked the can down the road w/r/t the proper SSA conversion that was already in the repository. Now I've updated that too.

I still need to add testing. I have spot-checked it on a few handpicked examples and it's going OK so far.


sampsyo:
Tests are all working and passing! And the Brench setup can now compare the two strategies.

PR #411, opened 2025-03-10
feat(benchmarks): Add fast inverse square root benchmark

ethanuppal:
This PR makes the following changes:

- A new mixed benchmark `fast-inverse-sqrt.bril` was added, using functions from `bitshift.bril`, that computes the approximate inverse square root of an IEEE floating point number. This benchmark uses the [bit casting extension I added to Bril a month ago](https://capra.cs.cornell.edu/bril/lang/bitcast.html). The current snapshot tests its correctness in computing $1/\sqrt{2}$.


ethanuppal:
Looks like CI is failing with this unrelated error:
image

sampsyo:
I think I've fixed the CI stuff! Could you please update the branch?

ethanuppal:
I've pushed an empty commit and the CI passes now!

sampsyo:
Thanks!

PR #412, opened 2025-03-11
Disable incorrect SSU enforcement in Brili reference interpreter

InnovativeInventor:
This line is incorrect. In particular, consider the following program as a counterexample:
```
@main() {
  one: int = const 1;
  set shadow one;
  main.cond: bool = const true;
  set cond main.cond;
.test:
  cond: bool = get;
  shadow: int = get;
  br cond .false .end;
.false:
  false.cond: bool = const false;
  set cond false.cond;
  jmp .test;
.end:
  print shadow;
}
```
I claim the following about the above program.
1) It is in SSA.
2) It follows the explicitly documented static and dynamic restrictions
   on `set`/`get`.
3) It always terminates with the output `1`.

However, when I run `brili`, I obtain the following output:
```
-> bril2json < repro.bril | brili
error: get without corresponding set for shadow
```

This commit disables the overly conservative enforcement mechanism. I suspect that SSU is best enforced via a static check rather than a dynamic runtime check.


InnovativeInventor:
@sampsyo CI seems to be failing on this PR and I'm not quite sure why. Perhaps I'm missing something, but the following error does not seem to be my fault:
(https://github.com/sampsyo/bril/actions/runs/13780468027/job/38537667879#step:17:2)
```
...
Error: Cache folder path is retrieved for pip but doesn't exist on disk: /home/runner/.cache/pip
...
```

ethanuppal:
I had the same issue on my PR - I think the general strategy on an open source project is to only ping the maintainer after a week of no reply

sampsyo:
Thank you!! I fixed the CI trouble and rebased this branch. You're right that this check was just wrong and me being overzealous; let's remove it.

PR #413, opened 2025-03-12
dont' remove from shadow env on get

Pat-Lafon:


sampsyo:
Awesome; very quick work here! Thank you!

PR #416, opened 2025-05-13
Add spawn/join concurrency extension to Bril

ananyagoenka:
This PR implements a lightweight thread‐based concurrency extension for Bril (closes #499).

## Highlights

1. **Specification**  
   - **`docs/extensions/concurrency.md`**: describes the new `thread` type, `spawn` (value‐op) and `join` (effect‐op), JSON syntax, and well‐formedness rules.  
   - **`docs/syntax.md`**: updated to list `spawn` and `join`, and the `thread` primitive.

2. **Parser Updates**  
   - **`tools/text/briltxt.py`**: Lark grammar now recognizes `spawn` and `join`.

3. **TypeScript Definitions**  
   - **`bril-ts/bril.ts`**:  
     - `PrimType` includes `"thread"`.  
     - `ValueOpCode` adds `"spawn"`, `EffectOpCode` adds `"join"`.  
     - `typeCheck` and `argCounts` updated.  
   - **`tests/concurrency_types_test.ts`**: simple `deno check` smoke test for new ops.

4. **Interpreter Runtime**  
   - **`src/brili.ts`**:  
     - Introduces `ThreadHandle`, a global `threadTable`, and `nextTid`.  
     - Implements `spawn` (copies args into a fresh `Env`, launches an async task, returns handle) and `join` (awaits the associated promise).  
     - New `--no-workers` flag for in-process async mode.

5. **Concurrency Test Suite**  
   - **`benchmarks/concurrency/`**:  
     - Passing tests (`*.bril` + `*.out`) covering correct behavior and errors.  
   - **`turnt.toml`**: adds `brili-conc` env to run the suite under `deno run -A brili.ts`.

6. **Micro-benchmarks & Performance**  
   - **`benchmarks/concurrency/perf/`**:  
     - `perf_par_sum_100k.bril`, `perf_matmul_split.bril`, `perf_big_par_sum_10M.bril`  
     - `run_perf.sh` measures sequential vs concurrent speed-up.

7. **Escape Analysis Pass**  
   - **`examples/escape.py`**: intraprocedural pass that reports thread-local `alloc`s.  
   - **`tools/README.md`**: how to run `escape.py` via `bril2json | python3 | deno run`.


bril-ts/brili_worker.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  // Build callee env, rehydrating Pointer arguments into Key-based Pointers
+  const callee = prog.functions.find((f) => f.name === func)!;
+  const env = new Map<bril.Ident, any>();
+  (callee.args || []).forEach((param, i) => {
+    let v = rawArgs[i];
+    // If the static type is a ptr<...>, re-wrap it
+    if (typeof param.type === "object" && Object.hasOwn(param.type, "ptr")) {
+      // v is { loc: { base: number, offset: number }, type: <Type> }
+      const raw = v as { loc: { base: number; offset: number } };
```
sampsyo, line 66: It looks like the root issue here is that `Key` objects, from the `brili` interpreter, can't be sent across threads? So the plan is to instead send a "plain" object with `number` fields across threads, and convert to/from `Key` instances when sending/receiving?

If that's the case, maybe we can make this simpler: let's just change `Key` so that it's simply the type `{ base: number; offset: number }`. That way, `Key`s will be directly portable across threads and we can avoid any conversion.

brili.ts
```diff
@@ -1,6 +1,72 @@
...
+  promise: Promise<void>;
+  resolve: () => void;
+  reject: (e: unknown) => void;
+  used: boolean;
+};
+
+const threadTable = new Map<number, ThreadRecord>();
+const workerPorts = new Map<number, Worker>();   // tid → worker
+let nextTid = 1;
+let globalHeap: Heap<Value>;
```
sampsyo, line 31: Let's avoid introducing global mutable state by moving these fields to the `State` object. That object is meant to encapsulate all the state for a running program, and things like the thread table can be considered part of that.

I am also guessing that this would eliminate the need for `globalHeap`, because we'd have access to `State.heap`.

docs/lang/concurrency.md
```diff
@@ -0,0 +1,112 @@
+Concurrency Extension
+=====================
+
+This extension brings *lightweight parallel threads* to Bril.
+
+*  **`spawn`** launches a new thread and returns an opaque handle.
+*  **`join`** blocks until a thread identified by such a handle finishes.
+*  **`thread`** is a new primitive type used for those handles.
```
sampsyo, line 8: This list seems a bit duplicative with the sections immediately below ("Primitive Types" and "Instructions"). Let's just keep one or the other, but not both—it would probably be easiest to just delete this list.

docs/lang/concurrency.md
```diff
@@ -0,0 +1,112 @@
...
+```jsonc
+{
+  "op":   "spawn",
+  "dest": "tid",
+  "type": "thread",
+  "funcs": ["func"],
+  "args":  ["arg1", "arg2", "..."]
+}
+
+* Creates a new thread that begins executing function @func.
```
sampsyo, line 41: Markdown backticks would be helpful here, and in other places where you refer to parts of the instruction:
```suggestion
* Creates a new thread that begins executing function `@func`.
```

PR #417, opened 2025-05-15
Add Flattened Representation/Interpreter for Bril

ngernest:
(Joint work w/ @samuelbreckenridge @katherinewu312)

This PR adds a new directory, `flat-bril`, which contains: 
- A [flattened](https://www.cs.cornell.edu/~asampson/blog/flattening.html) representation for the [Bril IR](https://capra.cs.cornell.edu/bril/)
- An interpreter that works natively over the flattened Bril representation
- Infrastructure for converting to/from Bril's canonical JSON format to the flattened representation

(More details in the [blogpost](https://www.cs.cornell.edu/courses/cs6120/2025sp/blog/flat-bril/) for this project.)

Code overview (in `flat-bril/`):
- [`main.rs`](./flat-bril/main.rs): Reads in a JSON Bril file from `stdin`
- [`flatten.rs`](./flat-bril/flatten.rs): Converts a JSON Bril file to a flattened instruction format 
- [`unflatten.rs`](./flat-bril/unflatten.rs): Converts a flattened Bril instruction back to JSON
- [`memfile.rs`](./flat-bril/memfile.rs): Serializes/De-serializes a flattened Bril file to/from disk
- [`interp.rs`](./flat-bril/interp.rs): Bril interpreter which works over the flattened Bril representation
- [`types.rs`](./flat-bril/flatten.rs): Type definitions & pretty-printers
- [`json_roundtrip.rs`](.flat-bril/json_round_trip.rs): Round-trip tests for converting from JSON -> flat format -> JSON

The docs have also been updated with a [page](./docs/tools/flat-bril.md) for `flat-bril`. 

To keep this PR small, I've omitted tests & benchmarks for `flat-bril` (these can be found in the [original repo](https://github.com/ngernest/flat-bril)).

### DIFF
--- a/docs/tools/flat-bril.md
+++ b/docs/tools/flat-bril.md
@@ -1,0 +1,39 @@
+# Flattened Representation 
+
+The `flat-bril` directory contains: 
+- A [flattened][adrian-blog] representation for Bril
+- An interpreter that works natively over the flattened Bril representation (`.fbril` files)
+  - Note: the interpreter currently only supports [core Bril](./lang/core.md).
+- Infrastructure for converting to/from Bril's canonical JSON format to the flattened representation
+
+Read [more about the implementation][blog], which is originally by Ernest Ng, Sam Breckenridge and Katherine Wu.
+
+**Caution**: The `flat-bril` interpreter is *not* meant to be a drop-in replacement for the reference Bril interpreters ([`brili`](./interp.md) and [`brilirs`](./brilirs.md)), as `flat-bril` only supports [core Bril](./lang/core.md). For most situations, we recommend using the reference Bril interpreter. 
+
+## Install
+To use `flat-bril` you will need to [install Rust][install-rust]. Use `echo $PATH` to check that `$HOME/.cargo/bin` is on your [path][path].
+
+In the `flat-bril` directory, install the tool with `cargo install --path .`.
+
+## Usage
+Run `flat-bril --help` to see all the supported command-line flags. 
+
+Here are some examples:      
+- To create a flattened Bril file `(call.fbril)` from an existing Bril file (eg. on `call.bril`):
+```bash
+$ bril2json < call.bril | flat-bril --filename call.fbril --fbril
+```
+- To interpret a flattened Bril file:
+```bash 
+$ flat-bril --filename call.fbril --interp
+```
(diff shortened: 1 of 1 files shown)


sampsyo:
Excellent work, @ngernest @samuelbreckenridge @katherinewu312! This all looks great! I'm sure there is plenty more experimentation that could be done based on this foundation, but I think it's in good shape to merge now. Maybe future projects will get inspired to build on this stuff.

One solid next step, if anyone is interested (and as you mentioned at the end of your comment above), would be to explore adding tests and benchmarking. For tests, there are two things to do:

* Check the interpreter correctness. Maybe this is as simple as adding a configuration to `test/interp/turnt.toml` and enabling the appropriate set of benchmarks in CI.
* Check round-tripping of the binary format. It seems like we can shoot for byte-for-byte identical JSON? As in, start with a text Bril file, then `bril2json < foo.bril` and `bril2json < foo.bril | flat-bril --json` (and perhaps a third command that involves actually writing to a file and reading it back), then the resulting JSON should be identical. I don't think this is quite true yet (there is some normalization w/r/t missing fields in the JSON vs. empty arrays), but it wouldn't be too hard to get there.

For benchmarking, we probably have some cross-cutting work to do here… there is some _basic_ support in the repo already for comparing the reference interpreter, brilirs, Brilift, and Brillvm, but it has kinda grown organically, one implementation at a time, and has never gotten the attention it deserves. (See [this summarize script](https://github.com/sampsyo/bril/blob/main/benchmarks/summarize.py), for instance.) It would be great to make this a bit more robust/flexible and incorporate your new interpreter as "just one more option" in the benchmarking setup.

Just for fun, a couple of next steps for optimizations occur to me to mention here, for posterity:

* As we discussed in person, enforcing uniqueness of names to avoid string comparison in the interpreter. This would be a fun way to exploit the flat format to do the interning for us. We'd probably want to build some generic machinery for a flat, interned set of strings (which we could reuse for variable names, opcodes, and types).
* "Streaming" conversion using some clever serde hacking (i.e., a [custom `Deserialize` implementation](https://serde.rs/impl-deserialize.html), or skipping serde altogether and switching to something like [struson](https://lib.rs/crates/struson). The idea would be to avoid materializing a `serde_json::Value` at all and directly construct the flat representation.

PR #418, opened 2025-05-17
Add Briloop

gerardogtn:
Add briloop to the bril repo!


bril-loop/briloop.jar
sampsyo, line 1: Please remove the jar file (the build product); we'll just keep the source files in the repository.

bril-loop/README.md
```diff
@@ -0,0 +1,29 @@
...
+* Java 21
+
+## Installation
+* Run `./install.sh` (requires sudo) will create a symbolic link in /usr/local/bin
+
+## Building
+* Run `./setup.sh` to download jar dependencies. 
+* Run `./build.sh` to compile the project. 
+* Run `./run.sh` to run the transformation.
+* Run `./release.sh` to generate the release jar.
```
sampsyo, line 24: Sorry if this is a silly question (I have not been in the JVM ecosystem for a long time), but is there a standard build tool (Gradle or Maven or [Amper](https://github.com/JetBrains/amper) or something) that we could use here to avoid custom shell scripting? A simple build setup could make this easier for people to use and maintain.

bril-loop/live.sh
sampsyo, line 1: Looks like this may have snuck in by mistake?

bril-txt/briltxt.py
```diff
@@ -262,11 +336,76 @@ def value_to_str(type, value):
         return str(value).lower()
 
 
-def instr_to_string(instr):
+def formatted_children(instrs, indentLevel):
+    formatted = map(lambda x: instr_to_string(x, indentLevel), instrs)
+    joined = '\n'.join(formatted)
+    return joined
```
sampsyo, line 342: A generator expression would be a bit shorter and easier to read:

```py
return '\n'.join(instr_to_string(i, indent_level) for i in instrs)
```

bril-txt/briltxt.py
```diff
@@ -262,11 +336,76 @@ def value_to_str(type, value):
         return str(value).lower()
 
 
-def instr_to_string(instr):
+def formatted_children(instrs, indentLevel):
```
sampsyo, line 339: Python variable names are traditionally camel_case, so `indent_level`.

brili.ts
```diff
@@ -131,6 +131,11 @@ const argCounts: { [key in bril.OpCode]: number | null } = {
   print: null, // Any number of arguments.
   br: 1,
   jmp: 0,
+  loop: 0, // ADDED
+  block: 0, // ADDED
+  if: 1, // ADDED
+  break: 0, // ADDED
+  continue: 0, // ADDED
```
sampsyo, line 138: We can dispense with the `ADDED` comments now. :)

brili.ts
```diff
@@ -330,8 +335,12 @@ type Action =
   | { "action": "end"; "ret": Value | null }
   | { "action": "speculate" }
   | { "action": "commit" }
-  | { "action": "abort"; "label": bril.Ident };
+  | { "action": "abort", "label": bril.Ident }
+  | { "action": "break", "level": number } // Break out of n loops.
+  | { "action": "continue", "level": number }; // Continue to next iteration of innermost loop.
```
sampsyo, line 340: The comment appears out-of-date, because `continue` also has a level parameter (so it's not just the innermost loop).

sampsyo (changes requested):
Very nice!! This is looking great, and quite self-contained.

I have a few high-level suggestions, both of which could be handled in separate PRs if you prefer:

* Do you have documentation for the control-flow language extension? Let's add that to the language extension docs.
* Do you have any tests for the relooper implementation? It would be great to include those too.

And here are some lower-level suggestions:

* Can we please name the top-level directory `reloop`? I've mostly used `bril-*` for the generic language libraries rather than for specific tools. Like, if we were to separate out the Bril language utilities from this relooper tool (not saying we should), those would go in a library called `bril-kt` or whatever.
* Let's add a documentation page for the tool itself. A good place to start would be by moving your existing `README.md` to `docs/tools/reloop.md`.
* I mentioned this in an inline comment as well, but any chance there's a simple way to adopt an existing Kotlin-friendly build tool (to avoid special-purpose shell scripts for installing dependencies and building the jar)?

PR #421, opened 2025-06-10
Flattening brilirs

Pat-Lafon:
Inspired by https://www.cs.cornell.edu/courses/cs6120/2025sp/blog/flat-bril/ I've slowly poked at something similar for `brilirs` though I have kept it internal(so no new input format, but also no zerocopy benefits).

```rust
pub enum FlatIR {
  Const {
    dest: VarIndex,
    value: Literal,
  },
  ZeroArity {
    op: ValueOps,
    dest: VarIndex,
  },
  UnaryArity {
    op: ValueOps,
    dest: VarIndex,
    arg: VarIndex,
  },
  BinaryArity {
    op: ValueOps,
    dest: VarIndex,
    arg0: VarIndex,
    arg1: VarIndex,
  },
  MultiArityCall {
    func: FuncIndex,
    dest: VarIndex,
    args: Vec<VarIndex>,
  },
  Nop,
  ...
}
```

This unexpectedly meant I needed to modify the typechecking step, so I moved that to before the cfg is created(a lot of mistakes were caught by the test suite!). Updating the interpreter part is mostly mechanical though I've slightly modified the loop to pass less arguments around.

I've done a small evaluation using the following command(swapping in each benchmark manually):
`hyperfine --warmup 3 --shell=none -L interp brilirs,brilirs_old '{interp} -f random_walk.json 2 5' --export-markdown random_walk.md`

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `brilirs -f ackermann.json 3 6` | 10.2 ± 2.2 | 9.5 | 46.7 | 1.00 |
| `brilirs_old -f ackermann.json 3 6` | 12.6 ± 0.8 | 11.8 | 21.9 | 1.23 ± 0.28 |
| `brilirs -f delannoy.json 8` | 27.6 ± 1.5 | 26.2 | 35.0 | 1.00 |
| `brilirs_old -f delannoy.json 8` | 32.0 ± 8.2 | 30.0 | 104.4 | 1.16 ± 0.30 |
| `brilirs -f leibniz.json` | 48.6 ± 2.3 | 47.1 | 64.5 | 1.00 |
| `brilirs_old -f leibniz.json` | 60.4 ± 1.0 | 59.5 | 65.0 | 1.24 ± 0.06 |
| `brilirs -f random_walk.json 2 5` | 4.4 ± 0.3 | 4.0 | 6.2 | 1.00 |
| `brilirs_old -f random_walk.json 2 5` | 4.9 ± 3.1 | 4.3 | 68.0 | 1.12 ± 0.71 |
| `brilirs -f mcccarthy91.json 15` | 2.1 ± 0.5 | 1.8 | 11.1 | 1.00 |
| `brilirs_old -f mcccarthy91.json 15` | 2.2 ± 1.2 | 1.8 | 36.8 | 1.01 ± 0.62 |
| `brilirs -f function_call.json 25` | 319.2 ± 2.7 | 317.0 | 325.5 | 1.01 ± 0.01 |
| `brilirs_old -f function_call.json 25` | 314.6 ± 2.6 | 311.1 | 320.5 | 1.00 |

The results are rather promising. A couple of non-trivial performance improvements on a biased set of benchmarks modulo my noisy laptop. mccarthy91 is probably too fast to get a meaningful result. function_call is probably dominated by call-stack manipulation which drowns everything else out.

It's a little hard to tell but the graph in the blog post doesn't seem to line up well with what I would expect in terms of relative values between benchmarks. If it's from this script then I think the graph mostly measures turnt overheads: https://github.com/ngernest/flat-bril/blob/24ca9993507da152d2bffe33eb39d18ca07909ba/bench.py#L21-L29

### DIFF
--- a/bril-rs/brillvm/src/llvm.rs
+++ b/bril-rs/brillvm/src/llvm.rs
@@ -1248,10 +1248,6 @@ pub fn create_module_from_program<'a>(
     let mut fresh = Fresh::new();
 
     // Add all functions to the module, initialize all variables in the heap, and setup for the second phase
-    #[expect(
-        clippy::needless_collect,
-        reason = "Important to collect, can't be done lazily because we need all functions to be loaded in before a call instruction of a function is processed."
-    )]
     let funcs: Vec<_> = functions
         .iter()
         .map(


brilirs/src/ir.rs
```diff
@@ -0,0 +1,419 @@
+use std::ops::{Add, Index};
+
+use bril_rs::{ConstOps, EffectOps, Instruction, Literal, ValueOps};
+use fxhash::FxHashMap;
+
+use crate::error::InterpError;
+
+/// A Newtype for function indexing
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct FuncIndex(pub usize);
```
sampsyo, line 10: Purely for the fun of it, one interesting experiment to run here would be to see what happens if you pick `u32` vs. `u64` for the various indices. Probably not a big impact, but it could be fun to measure just in case.

test/interp/dynamic/mem_dynamic_dispatch.bril
```diff
@@ -24,8 +24,8 @@
     ret;
 
 .call_my_print_2:
-    data : any = load args;
-    call @my_print_2 data;
+    data2 : any = load args;
+    call @my_print_2 data2;
```
sampsyo, line 28: Good catch… not entirely sure why the `brilck` checks didn't notice this type conflict with `data`. I'll look into this momentarily.

sampsyo (commented):
Nice; this is looking great already! Cool that most of the internal changes were mechanical.

test/interp/dynamic/mem_dynamic_dispatch.bril
sampsyo: Resolved here! Sorry that this will create a merge conflict, probably… https://github.com/sampsyo/bril/pull/422

Pat-Lafon:
> Purely for the fun of it, one interesting experiment to run here would be to see what happens if you pick `u32` vs. `u64` for the various indices. Probably not a big impact, but it could be fun to measure just in case.

I'm assuming this is to decrease the size of `FlatIR`?
The largest case is:
```rust
pub enum FlatIR {
  ...
  MultiArityCall {
    func: FuncIndex,
    dest: VarIndex,
    args: Vec<VarIndex>,
  },
  ...
}

const _: () = {
  assert!(40 == std::mem::size_of::<FlatIR>());
};
```
Changing to u32 actually didn't reduce the memory size surprisingly. I think this is because there are 2 other cases that have a `Vec<VarIndex>` field which somehow stops Rust from making the enum more compact?

I tried `pub type IndexType = u8;` which does reduce the size to `assert!(32 == std::mem::size_of::<FlatIR>());`. The results are impressive with across the board speedups for the `brilirs-small` version(with a fair amount of noise on my end).

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `./target/release/brilirs -f ackermann.json 3 6` | 10.0 ± 1.6 | 9.2 | 27.9 | 1.07 ± 0.18 |
| `../../brilirs-small/brilirs/target/release/brilirs -f ackermann.json 3 6` | 9.3 ± 0.6 | 8.7 | 15.0 | 1.00 |
| `./target/release/brilirs -f delannoy.json 8` | 27.5 ± 4.5 | 25.8 | 65.4 | 1.06 ± 0.21 |
| `../../brilirs-small/brilirs/target/release/brilirs -f delannoy.json 8` | 25.8 ± 3.0 | 24.1 | 49.8 | 1.00 |
| `./target/release/brilirs -f leibniz.json` | 48.0 ± 1.3 | 46.6 | 54.9 | 1.13 ± 0.09 |
| `../../brilirs-small/brilirs/target/release/brilirs -f leibniz.json` | 42.6 ± 3.1 | 41.3 | 66.9 | 1.00 |
| `./target/release/brilirs -f random_walk.json 2 5` | 4.2 ± 0.8 | 3.8 | 14.7 | 1.03 ± 0.31 |
| `../../brilirs-small/brilirs/target/release/brilirs -f random_walk.json 2 5` | 4.1 ± 1.0 | 3.7 | 27.2 | 1.00 |
| `./target/release/brilirs -f function_call.json 25` | 318.5 ± 3.5 | 312.6 | 323.0 | 1.07 ± 0.03 |
| `../../brilirs-small/brilirs/target/release/brilirs -f function_call.json 25` | 298.0 ± 8.6 | 288.0 | 316.8 | 1.00 |

(note I was cautious that for some reason the path length of the executable was to blame but these result replicate if ran from `brilirs-small/brilirs` directory instead)

I tried to go smaller(SmallVec, TinyVec, and Box<[VarIndex]>). I didn't observe any performance improvements from these attempts. The Box version was able to `assert!(24 == std::mem::size_of::<FlatIR>());` but without more performance to be found, maybe best not over-optimize on this. `pub type IndexType = u16;` is a nice sweet spot, 65k variables in a single function is probably sufficient.

Very cool!

sampsyo:
Very cool indeed! Yeah, 16-bit indices for local variables seems like a pretty sensible limit TBH. :)

PR #426, opened 2025-09-04
lesson2 feat: bechmark: add montecarlo simulation to calculate pi

jeffreyqdd:


sampsyo:
Very impressive! That's a pretty wild `to_float` implementation you have there. Kinda seems like we should probably add `int2float` and `float2int` operations to the floating-point extension… but anyway, this is more interesting/fun!

PR #430, opened 2025-09-08
new benchmark for use gradient descent to minimize f(w) = (w-t)^2

NingWang0123:
new benchmark for use gradient descent to minimize f(w) = (w-t)^2


NingWang0123:
Thank you so much for your help. I just fixed the missing types for the file.

NingWang0123:
Hi Prof. Sampson, I'm wondering if you could help me to kill the test that it is running right now? I found out that it will never stop since the editor auto upper the iteration i into I which make it running forever. I already pushed the new version of fixing this problem.

sampsyo:
Looks like the CI stuff eventually resolved itself!

NingWang0123:
It was automatically cancelled after  6 hrs running.

PR #439, opened 2025-09-09
Added a benchmark to print squares.

sunwookim028:
Added a benchmark to print squares.


sunwookim028:
I'll try to add something more interesting.

PR #444, opened 2025-09-10
Add Arithmetic Series Benchmark

adnan-armouti:
Tobi and I wanted to start off with something simple to get acquainted with Bril and Turnt - this benchmark implements the closed form solution to the arithmetic series problem.


benchmarks/core/arithmetic-series.bril
```diff
@@ -0,0 +1,9 @@
+@main(n: int) {
```
sampsyo, line 1: You'll need an `# ARGS: ...` comment here so that Turnt provides the argument to your `@main` function via the command line.

sampsyo (changes requested):
Looks like a perfectly reasonable idea overall! To make the tests work, you will need to supply an argument for your `@main` function. You can use `turnt -vp benchmarks/core/arithmetic-series.bril` to see this test failure locally.

sampsyo:
Great; all looking good now!

PR #445, opened 2025-09-10
add benchmark Kadane

zc579:


sampsyo (changes requested):
Hello! This generally looks headed in the right direction, but there are a few things to fix:

- Please add to the documentation. (See the assignment instructions, or other people's recent PRs, for what this should look like.)
- Please change your filename to be lower-case, to match all the other benchmarks.
- The profiling output doesn't match the count that you have recorded here. You can see the discrepancy with `turnt --diff benchmarks/mem/Kadane.bril`, or accept the new report with `--save`.

PR #447, opened 2025-09-10
Add mountain benchmark

tf-mac:
Adds a mountain benchmark, which determines if an integer is a mountain (intuitively, the digits go up then go down)


sampsyo:
Nice! This is a fun little idea.

PR #448, opened 2025-09-10
Added benchmarks - braille, square

sunwookim028:
Added two simple benchmarks:
- braille: converts the input integer into a Braille code format.
- square: computes squares of all integers from 1 up to the input integer.


benchmarks/core/braille.bril
```diff
@@ -0,0 +1,249 @@
...
+	six: int = const 6;
+	seven: int = const 7;
+	eight: int = const 8;
+	nine: int = const 9;
+	ten: int = const 10;
+	quotient: int = div input ten;
+	tmp: int = mul ten quotient;
+	last_digit: int = sub input tmp;
+	
+}
```
Pat-Lafon: I was updating my bril-to-llvm translator and it currently errors on this. Since the return type is int, but nothing is returned, I believe this is ill-typed(though apparently not caught by brilck).

Part of that might be because this code is dead?

sampsyo: Good point. I'm going to just delete this dead code, and then think on the other thread about how to check this statically!

PR #455, opened 2025-09-18
Rust 1.90 clippy

Pat-Lafon:


sampsyo:
Nicely done, as always!!

PR #470, opened 2026-05-17
Add installation instructions for running all tests README.md

techbelly:
When I followed the instructions in the README.md, I was unable to run all the tests without installing ts2bril and brilck. I've added instructions for installing those tools.


sampsyo:
Thanks! I rearranged some stuff in here for readability—namely, to indicate that you don't need these if you are only interested in specific tests.

